PR #27, opened 2018-03-14
design: add api design doc

bergwolf:
To fulfill the [kata design requirements](kata-design-requirements.md), and based on the disscusion on [Virtcontainers API extentions](https://docs.google.com/presentation/d/1dbGrD1h9cpuqAPooiEgtiwWDGCYhVPdatq7owsKHDEQ), [runtime API early sketch](https://github.com/kata-containers/runtime/pull/32) and [runtime API comparison](https://github.com/kata-containers/runtime/issues/46), the PR added the high level design of the kata runtime library API.

Fixes: #26

### DIFF
--- a/design/kata-api-design.md
+++ b/design/kata-api-design.md
@@ -1,0 +1,114 @@
+# Kata API Design
+To fulfill the [kata design requirements](kata-design-requirements.md), and based on the disscusion on [Virtcontainers API extentions](https://docs.google.com/presentation/d/1dbGrD1h9cpuqAPooiEgtiwWDGCYhVPdatq7owsKHDEQ), the kata runtime library features the following APIs:
+-  Sandbox based top API
+-  Storage and network hotplug API
+-  Plugin frameworks for external proprietary Kata runtime extensions
+-  Built-in shim and proxy types and capabilities
+
+## Sandbox Based API
+### Sandbox Management API
+|Name|Description|
+|---|---|
+|CreateSandbox(SandboxConfig)| Create and start a sandbox, and return the sandbox structure.|
+|FetchSandbox(ID)| Connect to an existing sandbox and return the sandbox structure.|
+|ListSandboxes()| List all existing standboxes with status. |
+
+### Sandbox Operation API
+|Name|Description|
+|---|---|
+|sandbox.Pause()| Pause the sandbox.|
+|sandbox.Resume()| Resume the paused sandbox.|
+|sandbox.Release()| Release a sandbox data structure, close connections to the agent, and quit any goroutines associated with the sandbox. Mostly used for daemon restart.|
+|sandbox.Delete()| Destroy the sandbox and remove all persistent metadata.|
+|sandbox.Status()| Get the status of the sandbox and containers.|
+|sandbox.Monitor()| Return a context handler for caller to monitor sandbox callbacks such as error termination.|
+|sandbox.CreateContainer()| Create new container in the sandbox.|
+|sandbox.DeleteContainer()| Delete a container from the sandbox.|
+|sandbox.StartContainer()| Start a container in the sandbox.|
+|sandbox.StatusContainer()| Get the status of a container in the sandbox.|
+|sandbox.EnterContainer()| Run a new process in a container.|
+|sandbox.WaitProcess()| Wait on a process to terminate.|
+### Sandbox Hotplug API
+|Name|Description|
+|---|---|
+|sandbox.AddStorage()| Add new storage to the sandbox.|
+|sandbox.AddNetwork()| Add new nic to the sandbox.|
+|sandbox.ListNetwork()| List all nics and their configurations in the sandbox.|
+|sandbox.UpdateNetwork()| Update the configuration of an existing network interface.|
+|sandbox.UpdateRoute()| Update the sandbox route table (e.g. for portmapping support).|
+
+### Sandbox Relay API
+|Name|Description|
+|---|---|
+|sandbox.WinsizeProcess(containerID, processID, Height, Width)|Relay TTY resize request to a process.|
+|sandbox.SignalProcess(containerID, processID, signalID, signalALL)| Relay a signal to a process or all processes in a container.|
+|sandbox.IOStream(containerID, processID)| Relay a process stdio. Return stdin/stdout/stderr pipes to the process stdin/stdout/stderr streams.|
+
+## Plugin framework for external proprietary Kata runtime extensions
+### Hypervisor plugin
+
+TBD.
+### Metadata storage plugin
+The metadata storage plugin controls where sandbox metadata is saved.
+All metadata storage plugins must implement the following API:
+
+|Name|Description|
+|---|---|
+|storage.Save(key, value)| Save a record.|
+|storage.Load(key)| Load a record.|
+|storage.Delete(key)| Delete a record.|
+
+Built-in implementations include:
+   -  Filesystem storage
+   -  LevelDB storage
+
+### VM Factory plugin
+The VM factory plugin controls how a sandbox factory creates new VMs.
+All VM factory plugins must implement following API:
+
+|Name|Description|
+|---|---|
+|VMFactory.NewVM(HypervisorConfig)|Create a new VM based on `HypervisorConfig`.|
+
+Built-in implementations include:
+
+|Name|Description|
+|---|---|
+|CreateNew()| Create brand new VM based on 'HypervisorConfig'.|
+|CreateFromTemplate()| Create new VM from template.|
+|CreateFromCache()| Create new VM from VM caches.|
+
+### Sandbox Creation Plugin Workflow
+![Sandbox Creation Plugin Workflow](https://raw.githubusercontent.com/bergwolf/raw-contents/master/kata/Kata-sandbox-creation.png "Sandbox Creation Plugin Workflow")
+
+### Sandbox Connection Plugin Workflow
+![Sandbox Connection Plugin Workflow](https://raw.githubusercontent.com/bergwolf/raw-contents/master/kata/Sandbox-Connection.png "Sandbox Connection Plugin Workflow")
+
+## Built-in Shim and Proxy Types and Capabilities
+### Built-in shim/proxy sandbox configurations
+-  Supported shim configurations:
+
+|Name|Description|
+|---|---|
+|noopshim|Do not start any shim process.|
+|ccshim| Start the cc-shim binary.|
+|katashim| Start the kata-shim binary.|
+|katashimbuiltin|No standalone shim process but shim functionality APIs are exported.|
+-  Supported proxy configurations:
+
+|Name|Description|
+|---|---|
+|noopProxy| a dummy proxy implementation of the proxy interface, only used for testing purpose.|
+|noProxy|generic implementation for any case where no actual proxy is needed.|
+|ccProxy|run ccProxy to proxy between runtime and agent.|
+|kataProxy|run kata-proxy to translate yamux connections between runtime and kata agent. |
+|kataProxyBuiltin| no standalone proxy process and connect to kata agent with internal yamux translation.|
+
+### Built-in Shim Capability
+Built-in shim capability is implemented by removing standalone shim process, and
+supporting the shim related APIs.
+
+### Built-in Proxy Capability
+Built-in proxy capability is achieved by removing standalone proxy process, and
+connecting to kata agent with a custom grpc dialer that is internal yamux translation.
+The behavior is enabled when proxy is configured as `kataProxyBuiltin`.


design/kata-api-design.md
```diff
@@ -0,0 +1,143 @@
...
+### Sandbox Management API
+-  CreateSandbox(ID, SandboxConfig)
+
+Create, start a sandbox and return the sandbox structure.
+-  ConnectSandbox(ID)
+
+Connect to an existing sandbox and return the sandbox structure.
+-  ListSandbox()
+
+List all existing standboxes with status.
```
jodh-intel: I'd be tempted to convert all these lists into tables for clarity:

| API | Description |
|--|--|
| `CreateSandbox(ID, SandboxConfig)` | Create, start a sandbox and return the sandbox structure
| `ConnectSandbox(ID)` | Connect to an existing sandbox and return the sandbox structure
| `ListSandbox()` | List all existing standboxes with status

design/kata-api-design.md
```diff
@@ -0,0 +1,143 @@
...
+Resume the paused sandbox.
+-  sandbox.Destroy()
+
+Destroy the sandbox and remove all persistent metadata.
+-  sandbox.Stats()
+
+Get the stats of the sandbox.
+-  sandbox.Done()
+
+Return a context handler that caller can monitor for sandbox callbacks such as
```
jodh-intel: s/that caller/the caller/ ?

design/kata-api-design.md
```diff
@@ -0,0 +1,143 @@
...
+Create, start a sandbox and return the sandbox structure.
+-  ConnectSandbox(ID)
+
+Connect to an existing sandbox and return the sandbox structure.
+-  ListSandbox()
+
+List all existing standboxes with status.
+
+### Sandbox Operation API
+-  sandbox.Release()
```
jodh-intel: What is the implication of calling this API? Does the sandbox continue to exist?

design/kata-api-design.md
```diff
@@ -0,0 +1,143 @@
...
+
+Add new nic to the sandbox.
+-  sandbox.UpdateRoute()
+
+### Sandbox Relay API
+Update the sandbox route table, e.g. for portmapping support.
+-  sandbox.ContainerTTYResize(containerID, processID, Height, Width)
+
+Relay TTY resize request to a process.
+-  sandbox.SignalProcess(containerID, processID, signalALL)
```
jodh-intel: Presumably this should be:

```
sandbox.SignalProcess(containerID, signalID, signalALL)
```

where `signalID` is either a name or an integer signal value?

design/kata-api-design.md
```diff
@@ -0,0 +1,143 @@
...
+
+Wait on a process.
+### Sandbox Hotplug API
+-  sandbox.AddStorage()
+
+Add new storage to the sandbox.
+-  sandbox.AddNic()
+
+Add new nic to the sandbox.
+-  sandbox.UpdateRoute()
```
jodh-intel: Missing description.

design/kata-api-design.md
```diff
@@ -0,0 +1,143 @@
...
+
+Run a new process in the sandbox.
+-  sandbox.WaitProcess()
+
+Wait on a process.
+### Sandbox Hotplug API
+-  sandbox.AddStorage()
+
+Add new storage to the sandbox.
+-  sandbox.AddNic()
```
jodh-intel: What about removing a NIC?

design/kata-api-design.md
```diff
@@ -0,0 +1,143 @@
+# Kata API Design
+To fulfill the [kata design requirements](kata-design-requirements.md), and based on the disscusion on [Virtcontainers API extentions](https://docs.google.com/presentation/d/1dbGrD1h9cpuqAPooiEgtiwWDGCYhVPdatq7owsKHDEQ), the kata runtime library features following APIs:
+-  Sandbox based top API
+-  Storage and network hotplug API
+-  Plugin frameworks for external proprietary Kata runtime extensions
+-  Built-in shim and proxy types and capabilities
+
+## Sandbox Based API
+### Sandbox Management API
+-  CreateSandbox(ID, SandboxConfig)
```
jodh-intel: Is there a reason none of the listed APIs show parameter types and return values? It is difficult to know how the API might behave without these imho.

design/kata-api-design.md
```diff
@@ -0,0 +1,143 @@
...
+   -  Supported Shim configurations:
+      -  noopshim
+      -  ccshim
+      -  katashim
+   -  Supported Proxy configurations:
+      -  noopProxy
+      -  noProxy
+      -  ccProxy
+      -  kataProxy
+      -  kataNoProxy
```
jodh-intel: I think we need descriptions for all of these as some are not obvious (for example what is the real difference between `noopProxy` and `noProxy`?)

design/kata-api-design.md
```diff
@@ -0,0 +1,143 @@
...
+
+`kataNoProxy` means no standalone proxy process and connect to kata agent with internal yamux translation.
+
+### Built-in Shim Capability
+Built-in shim capability is implemented by removing standalone shim process, and
+supporting the shim related APIs.
+
+### Built-in Proxy Capability
+Built-in proxy capability is archieved by removing standalone proxy process, and
```
jodh-intel: Typo: "achieved".

design/kata-api-design.md
```diff
@@ -0,0 +1,143 @@
...
+connect to kata agent with a custome grpc dialer that is internal yamux translation.
```
jodh-intel: Typo: "custom".

design/kata-api-design.md
bergwolf: good point!

design/kata-api-design.md
bergwolf: `noopProxy` and `noProxy` are existing types of proxy config in virtcontainers. IIUC `noopProxy` is to spawn a proxy process that does noting and it's for testing. `noProxy` means no proxy process at all.

And `kataNoProxy` means no proxy process + yamux grpc dialer, aka, proxy functionality without proxy process.

design/kata-api-design.md
bergwolf: Good catch! I missed the signalID. It should be `sandbox.SignalProcess(containerID, processID, signalID, signalALL)` where when `signalALL` is `true`, `processID` is ignored.

design/kata-api-design.md
bergwolf: ditto `ListNic` and `UpdateNic`. We need them to support similar functionality as `runv interface`. @WeiZhang555 could you please confirm if these are still **ALL** needed?

design/kata-api-design.md
bergwolf: Yes. The sandbox can continue running and CRI shims can reconnect to it with the `ConnectSandbox()` API.

bergwolf:
@jodh-intel thanks for reviewing! I'm delaying updating a bit to get more comments on the API set.

@sameo @sboeuf @WeiZhang555 @amshinde PTAL. I intentionally left blank most API parameters and all data structures to just focus on the API set and workflow. Those can be discussed in implementation PRs.

design/kata-api-design.md
jodh-intel: That's right. `noProxy` and `noopProxy` are explained here:

- https://github.com/containers/virtcontainers/blob/master/no_proxy.go#L23..L32
- https://github.com/containers/virtcontainers/blob/master/noop_proxy.go#L19..L20

 I think it would be useful to have this information in the document itself though.

bergwolf: You are right. I'll add explanation to each config when updating. Thanks!

design/kata-api-design.md
bergwolf: Sorry I missed this one. Yes, there is a reason -- I wanted to focus on the API set and workflow first. API parameter types/return values can be debatable for a long time and I'm deferring them to implementation PRs ;)

design/kata-api-design.md
```diff
@@ -0,0 +1,143 @@
...
+
+Add new storage to the sandbox.
+-  sandbox.AddNic()
+
+Add new nic to the sandbox.
+-  sandbox.UpdateRoute()
+
+### Sandbox Relay API
+Update the sandbox route table, e.g. for portmapping support.
+-  sandbox.ContainerTTYResize(containerID, processID, Height, Width)
```
sboeuf: This belongs to a specific process, and our caller should already have the `Process` handler in its hand:
```go
func (p *Process) WinsizeProcess(width, height int) error {}
```

design/kata-api-design.md
sboeuf: This belongs to a specific process, and should be handled this way:
```go
func (p *Process) SignalProcess(signal syscall.Signal, all bool) error {}
```

Please add more explanation here as I would like to understand when we actually need to signal a specific process.
Maybe this should be instead handled at the container level like this:
```go
func (c *Container) SignalContainer(signal syscall.Signal, all bool) error {}
```

design/kata-api-design.md
```diff
@@ -0,0 +1,143 @@
...
+
+Relay TTY resize request to a process.
+-  sandbox.SignalProcess(containerID, processID, signalALL)
+
+Relay a signal to a process or all processes in a container.
+-  sandbox.ProcessStdio(containerID, processID)
```
sboeuf: This belongs to a specific process, and should be handled this way:
```go
func (p *Process) IOProcess() ([]*os.File, error) {}
```
We need to explicitely define the order for the list of file descriptors. Something like:
```go
// Index list from IOProcess()
// 0 => STDIN
// 1 => STDOUT
// 2 => STDERR
```

design/kata-api-design.md
```diff
@@ -0,0 +1,143 @@
...
+Connect to an existing sandbox and return the sandbox structure.
+-  ListSandbox()
+
+List all existing standboxes with status.
+
+### Sandbox Operation API
+-  sandbox.Release()
+
+Release the sandbox connection associated with the sandbox structure.
+-  sandbox.Pause()
```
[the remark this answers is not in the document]
WeiZhang555: Better call this


func (p *Pod) Pause() error {}

instead of PausePod as you have a duplicate Pod word

design/kata-api-design.md
WeiZhang555: I think this is still necessary so far

design/kata-api-design.md
```diff
@@ -0,0 +1,143 @@
...
+
+Get the status of a container in the sandbox.
+-  sandbox.AddProcess()
+
+Run a new process in the sandbox.
+-  sandbox.WaitProcess()
+
+Wait on a process.
+### Sandbox Hotplug API
+-  sandbox.AddStorage()
```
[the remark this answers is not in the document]
bergwolf: A storage can be a rootfs as well. The virtcontainers `Volume` structure is too simple. We need a more complex `Storage` structure to describe different types of storages (file/dir/raw dev/rbd/nfs etc.). But IMO we should discuss the details in implementation PRs. This PR is about agreement on the API set and workflow.

design/kata-api-design.md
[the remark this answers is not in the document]
bergwolf: I thought the name is self-explaining enough... The use case is to hotplug the storage and nic. The expectation is to have the specified resource added to internal pod metadata and to the guest VM.

design/kata-api-design.md
bergwolf: `AddNetwork()` is about hotplug the nic device with a configuration for **the nic**. While `UpdateRoute()` is about updating the pod level routing table which can affect all nics.

design/kata-api-design.md
bergwolf: Again, then it will be a process level API. If we do want to export container and process level API, we need to get a consensus on it and have a way to get container and process structure from a sandbox. I know those exist in virtcontainers but we need to export them as well.

design/kata-api-design.md
bergwolf: We can spawn a process in a container. It is natural to support sending signals to the process we created rather than alway sending to all processes in a container.

design/kata-api-design.md
bergwolf: Yes, we need to. I just wanted to omit the implementation details in the PR.

bergwolf:
@sboeuf 
> You could modify Frakti and transform all the calls to ResizeTTY(), SignalProcess(), WaitProcess(), ..., by simply sending signals to the PID returned by the call to EnterContainer().

The API set is **NOT** the least requirements for CRI shims (frakti/hyperd) to work. It is to support how we expect CRI shims to work. IOW, it is a architecture design of frakti/hyperd to have no extra processes nor IPCs for agent communication. We do not expect that to change.

To be specific, what blockers do you have in mind that prevent virtcontainers from exporting these APIs?

sboeuf:
@bergwolf I don't have blockers in mind, I was just saying that getting the sandbox APIs ready will be much faster than having those process API (more work I think) ready, that's it.

design/kata-api-design.md
```diff
@@ -0,0 +1,143 @@
+# Kata API Design
+To fulfill the [kata design requirements](kata-design-requirements.md), and based on the disscusion on [Virtcontainers API extentions](https://docs.google.com/presentation/d/1dbGrD1h9cpuqAPooiEgtiwWDGCYhVPdatq7owsKHDEQ), the kata runtime library features following APIs:
+-  Sandbox based top API
+-  Storage and network hotplug API
+-  Plugin frameworks for external proprietary Kata runtime extensions
+-  Built-in shim and proxy types and capabilities
+
+## Sandbox Based API
```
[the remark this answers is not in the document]
sboeuf: I don't disagree with this but I am just trying to go step by step here, and I don't think the naming should be a huge priority as it will need to be changed in a bunch of different places in the code. Changing the whole naming could be part of a second step, otherwise we're also going to have to change what is already there.

design/kata-api-design.md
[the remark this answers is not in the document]
sboeuf: Yes I agree with you guys, as long as it's a method of `Pod`, we don't need the name to include pod.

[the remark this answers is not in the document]
sboeuf: @bergwolf 
Please update the document with:
```go
func (p *Pod) Pause() error {}
```

design/kata-api-design.md
```diff
@@ -0,0 +1,143 @@
...
+List all existing standboxes with status.
+
+### Sandbox Operation API
+-  sandbox.Release()
+
+Release the sandbox connection associated with the sandbox structure.
+-  sandbox.Pause()
+
+Pause the sandbox.
+-  sandbox.Resume()
```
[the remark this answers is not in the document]
sboeuf: @bergwolf 
Please update the document with:
```go
func (p *Pod) Resume() error {}
```

design/kata-api-design.md
```diff
@@ -0,0 +1,143 @@
...
+
+Resume the paused sandbox.
+-  sandbox.Destroy()
```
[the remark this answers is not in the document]
sboeuf: @bergwolf 
Please update the document with:
```go
func (p *Pod) Delete() error {}
```

design/kata-api-design.md
```diff
@@ -0,0 +1,143 @@
...
+
+Destroy the sandbox and remove all persistent metadata.
+-  sandbox.Stats()
```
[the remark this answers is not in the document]
sboeuf: @bergwolf 
Please update the document with:
```go
func (p *Pod) Status() (PodStatus, error) {}
```

design/kata-api-design.md
```diff
@@ -0,0 +1,143 @@
...
+
+Get the stats of the sandbox.
+-  sandbox.Done()
```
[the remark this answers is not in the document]
sboeuf: @bergwolf 
Please update the document with:
```go
func (p *Pod) Monitor() (<-PodMonitor, error) {}
```

design/kata-api-design.md
```diff
@@ -0,0 +1,143 @@
...
+
+Return a context handler that caller can monitor for sandbox callbacks such as
+error termination.
+-  sandbox.ListContainer()
```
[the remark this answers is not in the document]
sboeuf: Ok cool. Then I think that we can remove it for now, and in the meantime, we could try to assess if `pod.Status()` is taking too much time because it's doing some extra things not needed when we want only the list of containers.

design/kata-api-design.md
```diff
@@ -0,0 +1,143 @@
...
+
+List all containers in the sandbox.
+-  sandbox.CreateContainer()
```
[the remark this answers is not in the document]
sboeuf: Please update the doc.

design/kata-api-design.md
```diff
@@ -0,0 +1,143 @@
...
+
+Create new container in the sandbox.
+-  sandbox.DeleteContainer()
```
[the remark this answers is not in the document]
sboeuf: Yes that's my suggestion here. And actually, if you look at the current API, we're already returning/exporting `VCContainer` the same way we do with `VCPod`.
If you're okay with that, please update the doc with:
```go
func (c *Container) Delete() error {}
```

design/kata-api-design.md
```diff
@@ -0,0 +1,143 @@
...
+
+Delete a container from the sandbox.
+-  sandbox.StartContainer()
```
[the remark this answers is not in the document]
sboeuf: Yes, if you're okay with that, please update the doc with:
```go
func (c *Container) Start() error {}
```

design/kata-api-design.md
```diff
@@ -0,0 +1,143 @@
...
+
+Start a container in the sandbox.
+-  sandbox.StatusContainer()
```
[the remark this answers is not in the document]
sboeuf: If you're okay with that, please update the doc with:
```go
func (c *Container) Status() (ContainerStatus, error) {}
```

design/kata-api-design.md
```diff
@@ -0,0 +1,143 @@
...
+
+Get the status of a container in the sandbox.
+-  sandbox.AddProcess()
```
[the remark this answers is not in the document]
sboeuf: If you're okay with that, please update the doc with:
```go
func (c *Container) Enter() (VCProcess, error) {}
```

design/kata-api-design.md
```diff
@@ -0,0 +1,143 @@
...
+
+Run a new process in the sandbox.
+-  sandbox.WaitProcess()
```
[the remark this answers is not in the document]
sboeuf: Yes it is a process level API. While I was looking at all the needed functions, I realized it does not make sense to keep everything under a sandbox API, don't you think ?
```go
func (p *Process) Wait() error {}
```

design/kata-api-design.md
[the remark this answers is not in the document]
sboeuf: Yes we can discuss more details later, but I am not against changing `Volume` so that it can support this use case.

design/kata-api-design.md
sboeuf: Well we need inputs from other folks here, but it's really not a big deal to add functions like `FetchContainer()` or `FetchProcess()`, if we decide to go down this path.

design/kata-api-design.md
sboeuf: Then you don't need a release, you can just use a `FetchPod()` when it has been restarted.

sboeuf:
@bergwolf 
Ok so I have discussed with @sameo about this and we actually think that adding `Container` and `Process` API would increase the complexity of maintaining pointers consistency. So here is the sandbox API proposal based on your initial one, and based on the previous comments that we made:

```go
// Already exist
func CreatePod(podConfig PodConfig) (VCPod, error) {}

func FetchPod(podID string) (VCPod, error) {}

// Already exist.
func ListPod() ([]PodStatus, error) {}

func (p *Pod) Pause() error {}

func (p *Pod) Resume() error {}

func (p *Pod) Delete() error {}

func (p *Pod) Status() (PodStatus, error) {}

func (p *Pod) Monitor() (<-PodMonitor, error) {}

func (p *Pod) CreateContainer(ctrConfig ContainerConfig) (VCContainer, error) {}

func (p *Pod) DeleteContainer(ctrID string) error {}

func (p *Pod) StartContainer(ctrID string) error {}

func (p *Pod) StatusContainer(ctrID string) (ContainerStatus, error) {}

func (p *Pod) EnterContainer(ctrID string, cmd Cmd) (Process, error) {}

func (p *Pod) WaitProcess(ctrID, procID string) error {}

func (p *Pod) WinsizeProcess(ctrID, procID string, width, height int) error {}

func (p *Pod) SignalProcess(ctrID, procID string, signal syscall.Signal, all bool) error {}

func (p *Pod) IOProcess(ctrID, procID string) ([]*os.File, error) {}

func (p *Pod) AddStorage(storageConfig StorageConfig) error {}

func (p *Pod) AddNetwork(netConfig NetworkConfig) error {}
```

design/kata-api-design.md
sboeuf: Yes so this makes sense.

design/kata-api-design.md
sboeuf: I am happy either way about the naming.

design/kata-api-design.md
bergwolf: So we never need graceful shutdown? hmmm, it looks reasonable at first glance since all APIs are stateless and the kata agent does not quit on connection errors. @laijs WDYT?

design/kata-api-design.md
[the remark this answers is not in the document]
bergwolf: I agree that the API set can be implemented in many-step PRs. But the implementation steps do not block the API set we want to define/provide.

And about the pod->sandbox renaming, there is a good reason to do it and we'd better do it sooner than later.

design/kata-api-design.md
```diff
@@ -0,0 +1,143 @@
...
+-  Storage and network hotplug API
+-  Plugin frameworks for external proprietary Kata runtime extensions
+-  Built-in shim and proxy types and capabilities
+
+## Sandbox Based API
+### Sandbox Management API
+-  CreateSandbox(ID, SandboxConfig)
+
+Create, start a sandbox and return the sandbox structure.
+-  ConnectSandbox(ID)
```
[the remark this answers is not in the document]
bergwolf: @amshinde We want it to maintain an agent connection because otherwise most sandbox operation APIs would require the runtime to open a new connection to the agent, do the actual work and then close the connection, which is a waste of CPU/time. It does not need to be really stateful because multiple `FetchPod()` has the same effect and we do not need to actively close the connection. The only requirement is that `FetchPod()` being called before sandbox operations, which is the same in current VC workflow.

design/kata-api-design.md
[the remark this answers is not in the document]
bergwolf: Sure, we can change `Volume` to support it. It's just about naming.

design/kata-api-design.md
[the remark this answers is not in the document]
bergwolf: OK, I'll rename it.

bergwolf:
@sboeuf 
> I was just saying that getting the sandbox APIs ready will be much faster than having those process API (more work I think) ready, that's it.

Yes, I agree the sandbox APIs can be implemented faster. But the API design is not about single implementation change. It is to define what we want to provide, which will surely be implemented in multiple steps.

bergwolf:
@sboeuf 
> So here is the sandbox API proposal based on your initial one, and based on the previous comments that we made

I still think we need `sandbox.UpdateRoute()` and want to rename pod to sandbox to better match CRI semantics. I'll update the PR to reflect the discussion we've had so far.

design/kata-api-design.md
laijs: sandbox.Release() is for the graceful shutdown in case there are some long-living goroutines or other resources needed to be cleanup attached to the sandbox. The process should be able to release to a sandbox and attach it again. The first in-memory sandbox object should not have any leftover that interferes the later one.

runv's sandbox has several goroutines attached. I'm not sure about cc-runtime's.

bergwolf:
Thanks for reviewing!

Updated with suggestions from @jodh-intel  @sboeuf @WeiZhang555 @amshinde. PTAL. I still intentionally avoided most of the parameter types and return values. They can be discussed with implementation details.

I think we've agreed upon most of the API set. The remaining parts are:
1. pod -> sandbox rename
2. ListNetwork/UpdateNetwork/UpdateRoute
3. metadata storage plugin API and vm factory plugin API (I do not see any comments on them. Do you guys agree on what's proposed here?)

design/kata-api-design.md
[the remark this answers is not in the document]
sboeuf: Fair enough !

design/kata-api-design.md
```diff
@@ -0,0 +1,112 @@
...
+|sandbox.EnterContainer()| Run a new process in a container.|
+|sandbox.WaitProcess()| Wait on a process.|
+### Sandbox Hotplug API
+|Name|Description|
+|---|---|
+|sandbox.AddStorage()| Add new storage to the sandbox.|
+|sandbox.AddNetwork()| Add new nic to the sandbox.|
+|sandbox.ListNetwork()| List all nics and their configurations in the sandbox.|
+|sandbox.UpdateNetwork()| Update the configuration of an existing network interface.|
+|sandbox.UpdateRoute()| Update the sandbox route table, e.g. for portmapping support.|
```
sboeuf: I think this can be reduced to a more simple API:
- `AddStorage()`
- `AddNetwork()` (covering both `AddRoute` and `AddInterface`)
- `UpdateNetwork()` (covering both `UpdateRoute` and `UpdateInterface`)

Why do you need the list of networks with `ListNetwork()` ? CRI should already tell you what to add or what to update, right ?

design/kata-api-design.md
```diff
@@ -0,0 +1,112 @@
...
+|CreateSandbox(SandboxConfig)| Create, start a sandbox and return the sandbox structure.|
+|FetchSandbox(ID)| Connect to an existing sandbox and return the sandbox structure.|
+|ListSandbox()| List all existing standboxes with status. |
+
+### Sandbox Operation API
+|Name|Description|
+|---|---|
+|sandbox.Pause()| Pause the sandbox.|
+|sandbox.Resume()| Resume the paused sandbox.|
+|sandbox.Delete()| Destroy the sandbox and remove all persistent metadata.|
```
sboeuf: I think `Delete()` can be the one doing the `Release()` work you've been talking about, right ?

design/kata-api-design.md
```diff
@@ -0,0 +1,112 @@
...
+|Name|Description|
+|---|---|
+|sandbox.WinsizeProcess(containerID, processID, Height, Width)|Relay TTY resize request to a process.|
+|sandbox.SignalProcess(containerID, processID, signalID, signalALL)| Relay a signal to a process or all processes in a container.|
+|sandbox.IOStream(containerID, processID)| Relay a process stdio. Return stdin/stdout/stderr pipes to the process stdin/stdout/stderr streams.|
+
+## Plugin framework for external proprietary Kata runtime extensions
+### Hypervisor plugin
+
+TBD.
```
sboeuf: You can refer the POC here: https://github.com/kata-containers/runtime/pull/54

design/kata-api-design.md
```diff
@@ -0,0 +1,112 @@
...
+### Metadata storage plugin
+The metadata storage plugin controls where sandbox metadata is saved.
+All metadata storage plugins must implement following API:
```
sboeuf: Ok I need to understand why we need a plugin here. My understanding was that a plugin was useful in cases where we wanted proprietary implementations.
But if the point is to support a new type of storage, it should be directly implemented inside virtcontainers since the point of this library is to be reused by other projects, and we want to limit the potential duplication.

Last thing, you're defining some functions `Save/Load/Delete` but I'd like to understand why this needs to be defined through virtcontainers. Are you gonna make virtcontainers interacting with this ? If not, my understanding is that this could be directly implemented from the caller.

amshinde: Agree. Is this an API/Interface or a plugin? Same question for VM Factory plugin below.

design/kata-api-design.md
```diff
@@ -0,0 +1,112 @@
...
+|storage.Load(key)| Load a record.|
+|storage.Delete(key)| Delete a record.|
+
+Built-in implementations include:
+   -  Filesystem storage
+   -  LevelDB storage
+
+### VM Factory plugin
+The VM factory plugin controls how a sandbox factory creates new VMs.
+All VM factory plugins must implement following API:
```
sboeuf: Again, why are we talking about plugin here when I would expect (if my understanding is correct obviously), that we would implement a new hypervisor implementation `qemu_template.go` which is going to create a VM from template, and `qemu_cache.go` which is going to create a VM from cache. Both would implement the `hypervisor` interface from virtcontainers. And the `HypervisorConfig` would describe which implementation we need.
Is there a case that you think we won't be able to cover with such a design ?

design/kata-api-design.md
```diff
@@ -0,0 +1,112 @@
...
+
+Built-in implementations include:
+
+|Name|Description|
+|---|---|
+|CreateNew()| Create brand new VM based on HypervisorConfig|
+|CreateFromTemplate()| Create new VM from template.|
+|CreateFromCache()| Create new VM from VM caches.|
+
+### Sandbox Creation Plugin Workflow
```
sboeuf: The diagram looks fine, but should be updated with hypervisor implementations and not `plugins`. Also the title should be:
```
Sandbox Creation Workflow
```

design/kata-api-design.md
```diff
@@ -0,0 +1,112 @@
...
+![Sandbox Creation Plugin Workflow](https://raw.githubusercontent.com/bergwolf/raw-contents/master/kata/Kata-sandbox-creation.png "Sandbox Creation Plugin Workflow")
+
+### Sandbox Connection Plugin Workflow
```
sboeuf: Same here, the diagram should rely on virtcontainers new `storage` implementations and not plugins:
```
Sandbox Connection Workflow
```

design/kata-api-design.md
```diff
@@ -0,0 +1,112 @@
...
+|---|---|
+|noopProxy| a dummy proxy implementation of the proxy interface, only used for testing purpose.|
+|noProxy|generic implementation for any case where no actual proxy is needed.|
+|ccProxy|run ccProxy to proxy between runtime and agent.|
+|kataProxy|run kata-proxy to translate yamux connections between runtime and kata agent. |
+|kataNoProxy| no standalone proxy process and connect to kata agent with internal yamux translation.|
+
+### Built-in Shim Capability
+Built-in shim capability is implemented by removing standalone shim process, and
+supporting the shim related APIs.
```
sboeuf: Please add this would be a `kata_shim_builtin` implementation, given the fact that we already have `noop_shim`, `cc-shim` and `kata_shim`.

design/kata-api-design.md
```diff
@@ -0,0 +1,112 @@
...
+
+### Built-in Proxy Capability
+Built-in proxy capability is achieved by removing standalone proxy process, and
+connect to kata agent with a custom grpc dialer that is internal yamux translation.
+The behavior is enabled when proxy is configured as `kataNoProxy`.
```
sboeuf: Same here with `kata_proxy_builtin`.

sboeuf:
@bergwolf I have commented on the PR, but to summarize:
> 1. pod -> sandbox rename

Agreed.

> 2. sandbox.Release

This can be done through `sandbox.Delete()` I think.

> 3. ListNetwork/UpdateNetwork/UpdateRoute

Let's try to factorize a bit, and please give more context for the `ListNetwork()` use case.

> 4. metadata storage plugin API and vm factory plugin API (I do not see any comments on them. Do you guys agree on what's proposed here?)

I have a few comments about using the nomination `plugin` all the time. I'd like to make sure we're on the same page about those.

design/kata-api-design.md
bergwolf: It is to support CLI functionality like `runv network ls` as requested by @WeiZhang555.

And about combining API for Interface and route table, if you think about the arguments, they will have no overlap between each other. Are you suggesting that we use some kind of a union to combine them?

design/kata-api-design.md
bergwolf: `Delete()` is to destroy the sandbox. `Release()` is to release the outstanding agent connection and let the VM fly on its own. They are different functionalities.

design/kata-api-design.md
bergwolf: Yup. I think the hypervisor interface can be complex and we can discuss it in a separate PR.

design/kata-api-design.md
bergwolf: @sboeuf @amshinde please see the picture illustrating when these APIs are called **within** virtcontainers.

design/kata-api-design.md
bergwolf: The VM factory is on top of the hypervisor layer. It does not implement the `hypervisor` interface but can call in the `hypervisor` interface when necessary.

For example, if we want to create a VM directly without template or cache, we should:
1. create the VM by calling hypervisor `init()`/`createPod()`/`startPod()`

And if we want to create a new VM from template, we should:
1. clone a new VM from template
2. hotplug CPU and memory to the number as specified by `HypervisorConfig`

design/kata-api-design.md
bergwolf: @sboeuf the workflow is mainly to explain when VM Factory plugin and metadata plugin APIs are called.

design/kata-api-design.md
bergwolf: Again, the purpose is to focus on when/where the metadata plugin APIs are called during sandbox connection.

design/kata-api-design.md
bergwolf: Good point! Thanks!

bergwolf: Good point! Thanks!

design/kata-api-design.md
bergwolf: Will do!

design/kata-api-design.md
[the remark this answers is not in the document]
jodh-intel: I understand the reasoning but personally I really dislike the word "sandbox" as it's ambiguous:

- https://en.wikipedia.org/wiki/Sandbox_(computer_security)
- https://en.wikipedia.org/wiki/Sandbox_(software_development)

It might just be me, but when I think "sandbox" I invariably initially resolve that to mean "test area" before "security mechanism", which is wrong in this context of course.

Unless we want to go with a completely new noun like "vault", can I suggest that we add a brief definition of what we mean by "sandbox". This should mention that we're using the CRI terminology but also state that this is the equivalent of a k8s "pod" to avoid any confusion.

design/kata-api-design.md
jodh-intel: Since `CreateSandbox()` is performing multiple operations, internally this should be using some sort of transaction manager to ensure an atomic outcome:

- all operations completed successfully.
- an error occurred, so all *successful* operations are undone (to return the system to the state it was at before `CreateSandbox()` was called.

In summary: should we state that all APIs are atomic in operation?

design/kata-api-design.md
```diff
@@ -0,0 +1,112 @@
...
+|sandbox.Resume()| Resume the paused sandbox.|
+|sandbox.Delete()| Destroy the sandbox and remove all persistent metadata.|
+|sandbox.Status()| Get the status of the sandbox and containers.|
+|sandbox.Monitor()| Return a context handler for caller to monitor sandbox callbacks such as error termination.|
+|sandbox.CreateContainer()| Create new container in the sandbox.|
+|sandbox.DeleteContainer()| Delete a container from the sandbox.|
+|sandbox.StartContainer()| Start a container in the sandbox.|
+|sandbox.StatusContainer()| Get the status of a container in the sandbox.|
+|sandbox.EnterContainer()| Run a new process in a container.|
+|sandbox.WaitProcess()| Wait on a process.|
```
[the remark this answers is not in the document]
sboeuf: This means wait for a process to be "reparented" (aka terminate).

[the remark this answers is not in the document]
jodh-intel: I think we should state that in the description for clarity.

design/kata-api-design.md
bergwolf: @jodh-intel yes, we would want the sandbox operation APIs atomic to roll back any in-progress state if anything goes wrong.

design/kata-api-design.md
WeiZhang555: I might need to describe more about CLI interface of `runv/kata network add` especially scenarios

For current runv network support(as I know cc is similar, because I know runv better I'll take runv as example), runv will create a network namespace for runv-shim and docker will treat runv shim as a container. When a CNI plugin inserts a network interface into the "container"'s ns, runv can detect the interface with a nslistener daemon and try to create a tap and copy all configurations from veth to tap, this is how runv's CNI network worked initially.

That leads to some problems when runv is doing this "smart" things for user. First it's hard to detects and copy all the configurations, for example, if the CNI veth is put into an ovs bridge, runv will have to detects which ovs bridge the veth is in and its vlan port and other features(e.g. TSO feature is on or off), that brings lots of hard work to runv and make the "smart" things not that cool any more. What's worse, if user has a custom virtual switch that runv doesn't recognize(we happened to have one), it can do nothing, we have to break runv code and add more customized codes directly into runv.

So let's check how CNI works, a CNI plugin will insert an network interface and add it to some bridge and do more works, neither docker nor runc is aware of this. To imitate this, we contribute a `runv interface` command, if CNI plugin trys to put an network device into "VM"s namespace, it calls into `runv interface add` and the tap is inserted, and CNI plugin can do anything on host and runv doesn't need to care about anything. This is a more direct way than the "listener" and "detect" things.

For `ListNetwork` interface, CNI plugin can double check if the interface is configured correctly, even AddNetwork succeed we don't know if we truly get what we want, or maybe AddNetwork failed partly but we still get enough configuration to make CNI network work.

bergwolf: So `ListNetwork` is mostly for a CNI plugin to double check if the resulting network configuration is the same as expected, and thus it can fix any possible middle states, right?

bergwolf:
Updated per discussions we've had so far (here and on slack).

design/kata-api-design.md
```diff
@@ -0,0 +1,112 @@
...
+|---|---|
+|CreateNew()| Create brand new VM based on HypervisorConfig|
+|CreateFromTemplate()| Create new VM from template.|
+|CreateFromCache()| Create new VM from VM caches.|
+
+### Sandbox Creation Plugin Workflow
+![Sandbox Creation Plugin Workflow](https://raw.githubusercontent.com/bergwolf/raw-contents/master/kata/Kata-sandbox-creation.png "Sandbox Creation Plugin Workflow")
+
+### Sandbox Connection Plugin Workflow
+![Sandbox Connection Plugin Workflow](https://raw.githubusercontent.com/bergwolf/raw-contents/master/kata/Sandbox-Connection.png "Sandbox Connection Plugin Workflow")
```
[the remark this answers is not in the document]
bergwolf: Fixed. Thanks for spotting that!

design/kata-api-design.md
sboeuf: @WeiZhang555 Thank you so much for this very detailed explanation. This makes a lot of sense and I actually like the design :)

egernst: Much like how our discussion for gRPC went, wouldn’t it be better to return status and state if the add network fails?

bergwolf:
Thanks @sboeuf! I'll submit a following up PR to discuss the scope of v1.0 API targets once this is merged. I think most APIs in this document are required for v1.0 but you are right we can make plugin API exporting tentative. They are still required internally to support multiple builtin implementations imho.

design/kata-api-design.md
WeiZhang555: @bergwolf @egernst 

Yes, currently `ListNetwork` is only for double check, currently our CNI plugin will do a retry when `ListNetwork` returns unexpected result. Another usage is when end user(not CNI plugin) run an `AddNetwork` cmd, they generally would like to `ListNetwork` to check how many network interfaces he got in the end without calling into an `docker exec`.

For a CNI plugin, if AddNetwork can return final state of network interfaces, we don't need `ListNetwork`. But from an end user's perspective, a `ListNetwork` will be useful and friendly :-)

bergwolf: Sounds good to me. Thanks for the explanation @WeiZhang555

sboeuf: @jodh-intel @egernst PTAL and approve the PR if you're fine with it.

bergwolf:
ping @kata-containers/documentation

sboeuf:
@egernst @jodh-intel if this looks good to you, please let's move on and merge it :)

jodh-intel:
Aside from the "nit" about `ListSandbox()` being clearer if it was `ListSandboxes()` (since it's not just returning a single sandbox potentially)...

lgtm

bergwolf:
@jodh-intel changed to `ListSandboxes()`. Thanks!

Looks like we need one more +1 from @kata-containers/documentation to merge it. @egernst would you please take a look?

design/kata-api-design.md
```diff
@@ -0,0 +1,114 @@
...
+|CreateSandbox(SandboxConfig)| Create, start a sandbox and return the sandbox structure.|
+|FetchSandbox(ID)| Connect to an existing sandbox and return the sandbox structure.|
+|ListSandboxes()| List all existing standboxes with status. |
+
+### Sandbox Operation API
+|Name|Description|
+|---|---|
+|sandbox.Pause()| Pause the sandbox.|
+|sandbox.Resume()| Resume the paused sandbox.|
+|sandbox.Release()| Release a sandbox data structure, close connections to the agent and quit any goroutines associated with the sandbox. Mostly used for daemon restart.|
```
klynnrif: Adding in series comma: Release a sandbox data structure, close connections to the agent, and quit any goroutines associated with the sandbox. Mostly used for daemon restart.

design/kata-api-design.md
```diff
@@ -0,0 +1,114 @@
...
+-  Sandbox based top API
+-  Storage and network hotplug API
+-  Plugin frameworks for external proprietary Kata runtime extensions
+-  Built-in shim and proxy types and capabilities
+
+## Sandbox Based API
+### Sandbox Management API
+|Name|Description|
+|---|---|
+|CreateSandbox(SandboxConfig)| Create, start a sandbox and return the sandbox structure.|
```
klynnrif: Create and start a sandbox, and return the sandbox structure.

design/kata-api-design.md
```diff
@@ -0,0 +1,114 @@
...
+|sandbox.EnterContainer()| Run a new process in a container.|
+|sandbox.WaitProcess()| Wait on a process to terminate.|
+### Sandbox Hotplug API
+|Name|Description|
+|---|---|
+|sandbox.AddStorage()| Add new storage to the sandbox.|
+|sandbox.AddNetwork()| Add new nic to the sandbox.|
+|sandbox.ListNetwork()| List all nics and their configurations in the sandbox.|
+|sandbox.UpdateNetwork()| Update the configuration of an existing network interface.|
+|sandbox.UpdateRoute()| Update the sandbox route table, e.g. for portmapping support.|
```
klynnrif: Adding parenthesis for style consistency: Update the sandbox route table (e.g. for portmapping support).

design/kata-api-design.md
```diff
@@ -0,0 +1,114 @@
...
+|sandbox.SignalProcess(containerID, processID, signalID, signalALL)| Relay a signal to a process or all processes in a container.|
+|sandbox.IOStream(containerID, processID)| Relay a process stdio. Return stdin/stdout/stderr pipes to the process stdin/stdout/stderr streams.|
+
+## Plugin framework for external proprietary Kata runtime extensions
+### Hypervisor plugin
+
+TBD.
+### Metadata storage plugin
+The metadata storage plugin controls where sandbox metadata is saved.
+All metadata storage plugins must implement following API:
```
klynnrif: Suggested rewrite: All metadata storage plugins must implement the following API:

design/kata-api-design.md
```diff
@@ -0,0 +1,114 @@
...
+|storage.Load(key)| Load a record.|
+|storage.Delete(key)| Delete a record.|
+
+Built-in implementations include:
+   -  Filesystem storage
+   -  LevelDB storage
+
+### VM Factory plugin
+The VM factory plugin controls how a sandbox factory creates new VMs.
+All VM factory plugins must implement following API:
```
klynnrif: Suggested rewrite: All VM factory plugins must implement the following API:

design/kata-api-design.md
```diff
@@ -0,0 +1,114 @@
...
+
+|Name|Description|
+|---|---|
+|VMFactory.NewVM(HypervisorConfig)|Create a new VM based on `HypervisorConfig`.|
+
+Built-in implementations include:
+
+|Name|Description|
+|---|---|
+|CreateNew()| Create brand new VM based on HypervisorConfig|
```
klynnrif: Adding a period for consistency: Create brand new VM based on 'HypervisorConfig'.

design/kata-api-design.md
```diff
@@ -0,0 +1,114 @@
...
+
+### Sandbox Creation Plugin Workflow
+![Sandbox Creation Plugin Workflow](https://raw.githubusercontent.com/bergwolf/raw-contents/master/kata/Kata-sandbox-creation.png "Sandbox Creation Plugin Workflow")
+
+### Sandbox Connection Plugin Workflow
+![Sandbox Connection Plugin Workflow](https://raw.githubusercontent.com/bergwolf/raw-contents/master/kata/Sandbox-Connection.png "Sandbox Connection Plugin Workflow")
+
+## Built-in Shim and Proxy Types and Capabilities
+### Built-in shim/proxy sandbox configurations
+-  Supported Shim configurations:
```
klynnrif: -  Supported shim configurations:

design/kata-api-design.md
```diff
@@ -0,0 +1,114 @@
...
+
+|Name|Description|
+|---|---|
+|noopshim|do not start any shim process.|
+|ccshim| start the cc-shim binary.|
+|katashim| start the kata-shim binary.|
+|katashimbuiltin|no standalone shim process but shim functionality APIs are exported.|
+-  Supported Proxy configurations:
```
klynnrif: -  Supported proxy configurations:

design/kata-api-design.md
```diff
@@ -0,0 +1,114 @@
...
+### Sandbox Connection Plugin Workflow
+![Sandbox Connection Plugin Workflow](https://raw.githubusercontent.com/bergwolf/raw-contents/master/kata/Sandbox-Connection.png "Sandbox Connection Plugin Workflow")
+
+## Built-in Shim and Proxy Types and Capabilities
+### Built-in shim/proxy sandbox configurations
+-  Supported Shim configurations:
+
+|Name|Description|
+|---|---|
+|noopshim|do not start any shim process.|
```
klynnrif: Suggest capitalizing the beginning of each following description, for consistency.

design/kata-api-design.md
```diff
@@ -0,0 +1,114 @@
...
+|kataProxy|run kata-proxy to translate yamux connections between runtime and kata agent. |
+|kataProxyBuiltin| no standalone proxy process and connect to kata agent with internal yamux translation.|
+
+### Built-in Shim Capability
+Built-in shim capability is implemented by removing standalone shim process, and
+supporting the shim related APIs.
+
+### Built-in Proxy Capability
+Built-in proxy capability is achieved by removing standalone proxy process, and
+connect to kata agent with a custom grpc dialer that is internal yamux translation.
```
klynnrif: connecting to the kata agent  with a custom...

bergwolf:
Thanks for reviewing @klynnrif! Changes applied.

egernst:
@klynnrif - can you please take a look?

klynnrif:
Apologies @bergwolf @egernst I missed the earlier notification. Looks like you're set but let me know if not. Thanks!

PR #43, opened 2018-03-28
docs: Disable initrd in dev guide

### LINKED ISSUE #42: Update devguide for initrd
Now that https://github.com/kata-containers/runtime/pull/98 has landed, we need to update the dev guide for initrd handling.

jodh-intel:
Add details on how to disable the `initrd` config option to ensure the
base system as documented is functional.

Fixes #42.


jodh-intel:
Hi @klynnrif - ptal.

/cc @grahamwhaley, @egernst, @sboeuf - note that now https://github.com/kata-containers/runtime/pull/98 has landed, you must disable the `initrd` to get a working system.

egernst:
I suppose I should have questioned on the original issue #98 rather than here, but I'm wondering if it really makes sense to have this enabled by default in the first place?  Ie, as we build packages for Kata, will an initrd be used?

grahamwhaley:
/me crossed comments with @egernst 
Yeah, I didn't realise on #98 that things then built by hand would break (which broke on my local forked test repo for a bit as well...).
I'd like us to pick one default image type and have that as the default in the toml without having to `sed` around to un-break the system. I don't mind which. I feel I should ask if there is any reason it should not be an `initrd` - I think we'd need to hear the pro/cons of the image types really.

jodh-intel:
@grahamwhaley, @egernst - see my comments on this here:

- https://github.com/kata-containers/runtime/pull/98#issuecomment-376431565

egernst:
@jodh-intel  @grahamwhaley  - sure, but irrespective of how initrd is being set, we should set a sane default for kata-runtime configuration.  In this case, a default that doesn't break-by-default seems more sane at this point in time.  I'd rather see a configuration patch land than a documentation change to bypass the issue.  Thoughts?

grahamwhaley:
@egernst absolutely - I think we just need to agree what a sane working default is... which probably means choose initrd or not.

egernst:
Okay - I'll open an issue to have this conversation, and merge this PR to unblock users in the meantime.

Developer-Guide.md
```diff
@@ -62,6 +63,16 @@ $ sudo kata-runtime kata-check
 
 If your system is *not* able to run Kata Containers, the previous command will error and explain why.
 
+## Disable initrd
+
+Unless you plan to use an initial ramdisk (`initrd` or `initramfs`):
```
klynnrif: Suggested rewrite to make this active: "Unless you use an initial ramdisk (`initrd` or `initramfs`):"

Developer-Guide.md
```diff
@@ -62,6 +63,16 @@ $ sudo kata-runtime kata-check
...
+
+```
+$ sudo sed -i 's/^\(initrd.*\)/# \1/g' /usr/share/defaults/kata-containers/configuration.toml
+```
+
+If you do wish to use an initial ramdisk, see https://github.com/kata-containers/osbuilder.
```
klynnrif: If you use an initial ramdisk, see https://github.com/kata-containers/osbuilder.

jodh-intel:

PR #82, opened 2018-05-15
docs: Update dev guide for OBS packages

### LINKED ISSUE #80: Update dev guide for OBS installation packages
Add details to the developer guide explaining that the recommended way to install Kata is using the OBS binary packages.

Devs can then override particular components by building them from source.

jodh-intel:
Add details of how to install the packaged components to start off with
a working system. This involved splitting out part of the "Assumptions"
section into a new "Initial setup" section.

Fixes #80.

### DIFF
--- a/Developer-Guide.md
+++ b/Developer-Guide.md
@@ -1,5 +1,7 @@
 * [Warning](#warning)
 * [Assumptions](#assumptions)
+* [Initial setup](#initial-setup)
+* [Requirements to build individual components](#requirements-to-build-individual-components)
 * [Build and install the Kata Containers runtime](#build-and-install-the-kata-containers-runtime)
     * [Check hardware requirements](#check-hardware-requirements)
     * [Configure to use initrd or rootfs image](#configure-to-use-initrd-or-rootfs-image)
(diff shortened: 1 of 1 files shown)


jodh-intel:
Hi @klynnrif - ptal.

Developer-Guide.md
```diff
@@ -27,23 +29,32 @@
...
-
-    https://github.com/clearcontainers/runtime/wiki/Installation
+
+# Initial setup
+
+The recommended way to create a development environment is to first install the
+packaged versions of the Kata Containers components to create a working
+system:
+
+  * [Fedora*](install/fedora-installation-guide.md)
```
egernst: Will this create a mix of /usr/bin and /usr/local/bin on the system, though?  Should we just include directions on installing docker, qemu and VM artifacts?

Developer-Guide.md
```diff
@@ -27,23 +29,32 @@
...
+  * [Ubuntu*](install/ubuntu-installation-guide.md)
+
+The installation guide instructions will install all required Kata Containers
+components, plus Docker*, the hypervisor, and the Kata Containers image and
```
egernst: In general, I think all of our documentation assumes that all someone will ever want to do is setup Kata with Docker.  Its a good step, but I think we should be a bit more general, or at least include better install directions for Kubernetes development (this is an aside -- nothing we should address in this PR...).

Developer-Guide.md
jodh-intel: Yes - the official OBS binary for the runtime will be `/usr/bin/kata-runtime` and the developer version will be `/usr/local/bin/kata-runtime`. That's documented in the guide already (see https://github.com/kata-containers/documentation/blob/master/Developer-Guide.md#update-docker-configuration). We really don't want to encourage users to overwrite the packaged binaries, even on a dev box imho.

Developer-Guide.md
jodh-intel: Agreed. I think that's a post release concern ;)

jodh-intel:
Hi @klynnrif, @egernst et al - I've updated the branch so please could you take a fresh look? :)

PR #61, opened 2018-05-07
docs: Add ubuntu installation guide

GabyCT:
Add ubuntu installation guide for kata containers packages.

Fixes #60

Signed-off-by: Gabriela Cervantes <gabriela.cervantes.tellez@intel.com>

PR #83, opened 2018-05-15
install: Fix kata installation steps.

jcvenegas:
Fix kata installation URL.

### DIFF
--- a/install/fedora-installation-guide.md
+++ b/install/fedora-installation-guide.md
@@ -18,10 +18,11 @@ For more information on installing Docker please refer to the
 
 2. Install the Kata Containers components with the following commands:
 
+  **Note:** The repository redirects the download content to use `http`, be aware that this installation channel is not secure.
 ```bash
 $ source /etc/os-release
 $ sudo -E VERSION_ID=$VERSION_ID dnf config-manager --add-repo \
-https://download.opensuse.org/repositories/home:/katacontainers:/release/Fedora\_$VERSION_ID/home:katacontainers:release.repo
+http://download.opensuse.org/repositories/home:/katacontainers:/release/Fedora\_$VERSION_ID/home:katacontainers:release.repo
 $ sudo -E dnf -y install kata-runtime kata-proxy kata-shim
 ```
 
--- a/install/ubuntu-installation-guide.md
+++ b/install/ubuntu-installation-guide.md
@@ -22,9 +22,10 @@ For more information on installing Docker please refer to the
 
 2. Install the Kata Containers components with the following commands:
 
+  **Note:** The repository is downloading content using `http`, be aware that this installation channel is not secure.
 ```bash
-$ sudo sh -c "echo 'deb https://download.opensuse.org/repositories/home:/katacontainers:/release/xUbuntu_$(lsb_release -rs)/ /' >> /etc/apt/sources.list.d/kata-containers.list"
-$ curl -sL  https://download.opensuse.org/repositories/home:/katacontainers:/release/xUbuntu_$(lsb_release -rs)/Release.key | sudo apt-key add -
+$ sudo sh -c "echo 'deb http://download.opensuse.org/repositories/home:/katacontainers:/release/xUbuntu_$(lsb_release -rs)/ /' >> /etc/apt/sources.list.d/kata-containers.list"
+$ curl -sL  http://download.opensuse.org/repositories/home:/katacontainers:/release/xUbuntu_$(lsb_release -rs)/Release.key | sudo apt-key add -
 $ sudo -E apt-get update
 $ sudo -E apt-get -y install kata-runtime kata-proxy kata-shim
 ```


jodh-intel:
Hi @jcvenegas - I'm afraid we don't want to do this. Everything should be using https now. I think you've been hit by the same proxy problem as @eadamsintel had on https://github.com/kata-containers/documentation/pull/61.

jodh-intel:
(ftr, it works perfectly using https without proxies ;)

egernst:
@jodh-intel I ran into this issue on an external (cloud) machine.  No proxy.

jodh-intel:
It sounds like this is a package version issue. I tested with Ubuntu 17.10. We *really* want to avoid having to document an insecure installation method using http though.

chavafg:
`https` does not work for me when installing under Ubuntu 16.04.

egernst:
@jcvenegas I think this is xenial specific.  Perhaps just a short term w/a necessary while we still package Xenial?

jodh-intel:
I've recreated the problem on Ubuntu 16.04. In fact, you can simulate it in any environment like this by telling `curl` to allow redirects but **only** to another https url:

```
$ curl -sSL --proto \=https --proto-redir \=https -O https://download.opensuse.org/repositories/home:/katacontainers:/release/xUbuntu_16.04/./amd64/qemu-vanilla_2.11+git.e3050471ff-35_amd64.deb  
curl: (1) Protocol "http" not supported or disabled in libcurl
```

The problem is that the opensuse download site appears to be redirecting the incoming https request to an http download url, even though the connection originally came in via https. If correct, that seems rather broken to me.

What's odd is that the opensuse download site *can* service https downloads fwics. Adding curl debug (`-vvv`) shows that after the https -> http redirect, the actual file that will be downloaded is:

- http://downloadcontent.opensuse.org/repositories/home:/katacontainers:/release/xUbuntu_16.04/amd64/qemu-vanilla_2.11+git.e3050471ff-35_amd64.deb

And yet, you *can* download that file over https!:

```
$ curl -L --proto \=https --proto-redir \=https -O https://downloadcontent.opensuse.org/repositories/home:/katacontainers:/release/xUbuntu_16.04/amd64/qemu-vanilla_2.11+git.e3050471ff-35_amd64.deb
```

Hence, I think this issue can be summarised as "Ubuntu 16.04 is dtrt wrt security but the opensuse server *appears* to be 'not configured optimally', atleast for libcurl clients ;-)".

Unless opensuse change their server configuration, I think we are indeed going to be forced into changing the urls to "http" for *two* reasons:

1) To allow the Ubuntu 16.04 to actually work.
2) To make it clear to *all* users installing Kata that this is an insecure download
    (since even if you specify https urls, they get rewritten to http silently by newer distro versions).

/cc @cseader.

egernst:
@jodh-intel based on this; this PR shouldn't be "do not merge" at this point, right?

jodh-intel:
Well, imho we need to:

- also revert the fedora install doc to specify http since if you see my comments above even though OBS supports https URLs, they appear to redirect to http. That's bad from a security standpoint.
- add warnings to both docs that the installation method is insecure due to OBS's servers.

jcvenegas:
hum odd travis fail.
```
$ ruby --version
ruby 2.4.1p111 (2017-03-22 revision 58053) [x86_64-linux]
$ rvm --version
rvm 1.29.3 (latest) by Michal Papis, Piotr Kuczynski, Wayne E. Seguin [https://rvm.io]
$ bundle --version
Bundler version 1.16.2
$ gem --version
2.7.7
0.16s$ rake
rake aborted!
No Rakefile found (looking for: rakefile, Rakefile, rakefile.rb, Rakefile.rb)
/home/travis/.rvm/gems/ruby-2.4.1@global/gems/rake-12.3.0/exe/rake:27:in `<top (required)>'
(See full trace by running task with --trace)
```

Not a blocker

jodh-intel:
Hi @vrothberg - could you provide any further details on the https -> http issue mentioned above (https://github.com/kata-containers/documentation/pull/83#issuecomment-389474828)?

vrothberg:
@jodh-intel Thanks for reaching out! I will make sure our build team has a look at the issue.

jodh-intel:
@vrothberg - no problem and thanks! :smile:

jodh-intel:
@vrothberg, @adrianschroeter - I've checked this morning and the https -> http issue seems to be mostly resolved so thanks for fixing this this!

However, a random sampling of downloads for https://download.opensuse.org/repositories/home:/katacontainers:/release/ shows that the only distros where we are unable to download entirely via https are... OpenSUSE and SLES:

```
$  curl -sS -L --proto \=https --proto-redir \=https -O https://download.opensuse.org/repositories/home:/katacontainers:/release/openSUSE_Leap_42.3/x86_64/kata-linux-container-debug-4.14.51.1-132.1.x86_64.rpm
curl: (1) Protocol "http" not supported or disabled in libcurl
$ curl -sS -L --proto \=https --proto-redir \=https -O https://download.opensuse.org/repositories/home:/katacontainers:/release/SLE_12_SP3/x86_64/kata-proxy-1.1.0+git.8a305e5-30.1.x86_64.rpm
curl: (1) Protocol "http" not supported or disabled in libcurl
```

Could you tweak the settings for those repos too please?

marcov:
Hi @jodh-intel,  not all the mirrors in the mirror infrastructure supports HTTPS. That's also why the `.repo` includes URL with HTTP and not HTTPS.

But as I explained [here](https://github.com/kata-containers/documentation/issues/212#issuecomment-425004396) HTTP is not an issue, and in my view the warning should be removed from the doc pages.

(Speaking for SUSE) Probably you can safely use HTTPS as the main repo URL, but individual packages will still be downloaded using HTTP.

PR #137, opened 2018-06-05
Ubunut/Kata: upgrade from clear container to kata container

JerryOpenix:
Hit below error when upgrading on Ubuntu:
    The following packages have unmet dependencies:
    kata-runtime : Depends: kata-ksm-throttler (>= 1.0.0+git.422c7f7-28)
    but it is not going to be installed
Fix it by:
    sudo dpkg -P cc-ksm-throttler

Signed-off-by: Liu Changcheng <changcheng.liu@intel.com>

### DIFF
--- a/Upgrading.md
+++ b/Upgrading.md
@@ -89,14 +89,14 @@ Kata Containers configuration.
 ### Fedora
 
 ```
-$ sudo -E dnf remove cc-runtime\* cc-proxy\* cc-shim\* linux-container clear-containers-image qemu-lite
+$ sudo -E dnf remove cc-runtime\* cc-proxy\* cc-shim\* linux-container clear-containers-image qemu-lite cc-ksm-throttler
 $ sudo rm /etc/yum.repos.d/home:clearcontainers:clear-containers-3.repo
 ```
 
 ### Ubuntu
 
 ```
-$ sudo apt-get purge cc-runtime\* cc-proxy\* cc-shim\* linux-container clear-containers-image qemu-lite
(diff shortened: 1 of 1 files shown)


jodh-intel:
Hi @JerryOpenix - thanks for raising. Although the PR adds the command you were told to run, I'd rather we handle this more directly by simply adding `cc-ksm-throttler` to the `sudo apt-get purge ...` line above. Can you add the `cc-ksm-throttler` package to the `sudo -E dnf remove ...` Fedora section too?

> **Aside:**
>
> I've just noticed there is an inconsistency with the document - we specify
> `sudo -E` for Fedora but only `sudo` for Ubuntu. It might be worth adding the `-E` flag for Ubuntu
> on a separate commit as that is the standard `sudo` invocation we use in our docs.

On to why the CI is failing...

Our process requires that commits conform to a certain format. Travis is failing as a result and if you look at the Travis log it points you at this document which explains the format of commits our CI tooling checks for:

- https://github.com/kata-containers/community/blob/master/CONTRIBUTING.md#patch-format

Basically, your commit needs to add:

- A signed-off-by comment (which you've already done).
- A "subsystem" tag in the summary line.
- A "Fixes" comment referencing a github issue the PR is fixing (I've raised #138 for you to use).

Hence,  if you change your commit to something like the following, the CI should be happy:

```
docs: Remove throttler in upgrading doc

Updated the `Upgrading.md` document to ensure users remove the Clear Containers throttler package to avoid conflicts with the Kata Container equivalent.

Fixes #138.

Signed-off-by: Liu Changcheng <changcheng.liu@intel.com>
```

devimc:
lgtm

jodh-intel:
lgtm

PR #135, opened 2018-06-01
docs: Update documentation to provide Kubernetes alternative to Docker

sboeuf:
Most of the documentation assumes the user will use Kata Containers with Docker. But this assumption is wrong most of the time, and it is important to provide some documentation on how Kata Containers integrate with Kubernetes.

This PR first updates some inconsistencies in the doc before to actually provide some explanations on how to setup Kubernetes with Kata Containers. Last commit is about factorizing the docs provided for each distribution, and make them point to the newly added doc.

Fixes #134

PR #141, opened 2018-06-06
CI: test installation guides

### LINKED ISSUE #132: Add install testing CI test
Once we have jenkins setup for this repo, create a CI test that executes the install docs [*] when they change to ensure we haven't borked them.

---

[*] - by calling https://github.com/kata-containers/tests/blob/master/cmd/kata-manager/kata-manager.sh with the `install-packages` command.

jodh-intel:
Moved static tests to CI setup script and added a new CI test to execute
all install guides if any one changes.

Fixes #132.

### DIFF
--- a/.ci/test-install-docs.sh
+++ b/.ci/test-install-docs.sh
@@ -1,0 +1,86 @@
+#!/bin/bash
+#
+# Copyright (c) 2018 Intel Corporation
+#
+# SPDX-License-Identifier: Apache-2.0
+
+set -e
+
+info()
+{
+	msg="$*"
+	echo "INFO: $msg"
+}
+
+# Detect if any installation guides changed. If so, run the
+# kata manager to "execute" the install guide to ensure the
+# commands it specified result in a working system.
+check_install_guides()
+{
+	local -r install_guide_suffix="-installation-guide.md"
+
+	# List of filters used to restrict the types of file changes.
+	# See git-diff-tree(1) for further info.
+	local filters=""
+
+	# Added file
+	filters+="A"
+
+	# Copied file
+	filters+="C"
+
+	# Modified file
+	filters+="M"
+
+	# Renamed file
+	filters+="R"
+
+	# Unmerged (U) and Unknown (X) files. These particular filters
+	# shouldn't be necessary but just in case...
+	filters+="UX"
+
+	# List of changed files
+	local files=$(git diff-tree \
+		--name-only \
+		--no-commit-id \
+		--diff-filter="${filters}" \
+		-r \
+		origin/master HEAD || true)
+
+	# No files were changed
+	[ -z "$files" ] && return
+
+	changed=$(echo "$files" | grep "install/.*${install_guide_suffix}$" || true)
+
+	[ -z "$changed" ] && info "No install guides modified" && return
+
+	info "Found modified install guides: $changed"
+
+	# Regardless of which distro install guide(s) were changed, we test
+	# them all where possible.
+
+	local -r GOPATH=$(go env GOPATH)
+	[ -z "$GOPATH" ] && die "cannot determine GOPATH"
+
+	local -r mgr="${GOPATH}/src/github.com/kata-containers/tests/cmd/kata-manager/kata-manager.sh"
+
+	[ ! -e "$GOPATH" ] && die "cannot find $mgr"
+
+	source /etc/os-release
+
+	info "Installing system from the $ID install guide"
+
+	$mgr install-packages
+
+	[ -n "$TRAVIS" ] && info "Not creating container as Travis lacks VT-x" && return 
+
+	$mgr configure-image
+	$mgr enable-debug
+
+	msg="INFO: Successfully tested install guide for distro '$ID' $VERSION"
+
+	# Perform a basic test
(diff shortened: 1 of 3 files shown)


jodh-intel:
Blocked on https://github.com/kata-containers/tests/pull/384.

jodh-intel:
@chavafg, @grahamwhaley, @egernst, @gnawux - this is a particularly interesting PR which I'd *really* like to get landed :)

Note the blocking PR and my cunning plan:

- https://github.com/kata-containers/documentation/pull/135#discussion_r193356040

.ci/test-install-docs.sh
```diff
@@ -0,0 +1,87 @@
...
+	$mgr configure-image
+	$mgr enable-debug
+
+	msg="INFO: Successfully tested install guide for distro '$ID' $VERSION"
+
+	# Perform a basic test
+	sudo docker run --rm -i --runtime "kata-runtime" busybox echo "$msg"
+}
+
+main()
```
grahamwhaley: the whole main calls main calls the only function in the file feels slightly redundant :-)

.ci/test-install-docs.sh
```diff
@@ -0,0 +1,87 @@
...
+	# List of changed files
+	local files=$(git diff-tree \
+		--name-only \
+		--no-commit-id \
+		--diff-filter="${filters}" \
+		-r \
+		origin/master HEAD || true)
+
+	# No files were changed
+	[ -z "$files" ] && return
```
grahamwhaley: As this is a CI - maybe a message here for the logs noting we finished as there was nothing to do (rather than a silent quit, which I predict some day somebody will have to go 'debug' :-) )

.ci/test-install-docs.sh
```diff
@@ -0,0 +1,87 @@
...
+
+	local -r mgr="${GOPATH}/src/github.com/kata-containers/tests/cmd/kata-manager/kata-manager.sh"
+
+	[ ! -e "$GOPATH" ] && die "cannot find $mgr"
+
+	source /etc/os-release
+
+	echo "INFO: Installing system from the $ID install guide"
+
+	$mgr install-packages
```
grahamwhaley: I'm a little confused how `$mgr` relates to `$ID`, and then if/how that has any affect/interaction/relationship to `$changed` - maybe a comment here giving a quick overview of the relationship would be good (as it is not clear to me from just the code that if `$changed` does not include the file relevant to `$ID` - do we do anything useful, and would we check the file that changed iyswim).
I know, it is partly my ignorance for not following the details of how the auto-docs scripts work - but I'm sure I'm not the only one ;-)

.ci/test-install-docs.sh
```diff
@@ -0,0 +1,87 @@
...
+	# Travis doesn't support VT-x
+	[ -n "$TRAVIS" ] && return
+
+	$mgr configure-image
+	$mgr enable-debug
+
+	msg="INFO: Successfully tested install guide for distro '$ID' $VERSION"
+
+	# Perform a basic test
+	sudo docker run --rm -i --runtime "kata-runtime" busybox echo "$msg"
```
grahamwhaley: I'm not sure we actually need the `sudo` do we (@chavafg) - I thought our default users would be in the docker group?

.ci/test-install-docs.sh
jodh-intel: Fixed.

.ci/test-install-docs.sh
jodh-intel: Fixed.

.ci/test-install-docs.sh
jodh-intel: I'll leave it for now as it's more likely to work this way but will change if @chavafg says it's redundant.

.ci/test-install-docs.sh
jodh-intel: I'm not quite clear on what the problem is I'm afraid?

There are two comments in the test that explain that if *any* install guide changes, we re-test *all* the install guides (why not? :) The script is simply using `$ID` for informational purposes - the `$mgr` script will dtrt and install the packages for the distro the script is called on (or error if it cannot find an install doc for the current distro).

grahamwhaley: I know it says 'all the scripts', and then the script itself says `INFO: Installing system from the $ID install guide`. I just didn't, obviously, see how 'all' fitted in with that specific looking echo. np - as I said, not obvious to me as I don't know exactly how this ties into the scripts it is calling.

jodh-intel: Oh right. What I was trying to get as was, since this is a CI script, it will run on 'n' distros and if the install guide for distro 'a' is modified, we'll still re-check all 'n' distros (by running this new CI script once on each of the 'n' distros our CI is configured for). This seems reasonable given the jobs run in parallel and it doesn't hurt for us to re-test as much as possible on all distros given package versions (and their dependent package versions) may change frequently.

chavafg:
lgtm

.ci/test-install-docs.sh
chavafg: yes, we have our default user in the docker group, but for me it is ok to leave it like this.

caoruidong:
lgtm

PR #139, opened 2018-06-05
check if hardwarte is capable of running Kata Containers

n3wt0nSAN:
docs: Add a section "Hardware requirements" on Ubuntu user guide.

Fixes [#373](https://github.com/kata-containers/runtime/issues/373).

Signed-off-by: Newton Jose <josephnewton001@gmail.com

PR #143, opened 2018-06-06
reference to hardware requirements

### LINKED ISSUE #373: Decide on docker version strategy
Docker removed devicemapper in 18.09 which complicates life for many (see https://github.com/kata-containers/runtime/pull/1112).

Our [install docs](https://github.com/kata-containers/documentation/tree/master/install) setup the docker repo and then help users *install the latest version of docker*. Given the above, that's a potential problem so we need to find a non-gross way of updating the install docs to install docker 18.06.

The approach needs to:

- be simple.
- work with [`kata-manager`](https://github.com/kata-containers/tests/tree/master/cmd/kata-manager).

Since the docker package will get auto-upgraded from 18.06 -> 18.09 when the user upgrades their installed packages, we're basically going to have to `apt-mark hold` docker I think:

- https://github.com/kata-containers/runtime/issues/926#issuecomment-440689612

We'll also have to find the appropriate distro rune for each other install doc and then remove it if/when docker can be upgraded once again. And add a "note" explaining what the extra step is for.

/cc @egernst.

n3wt0nSAN:
docs: Add a section "hardware requirements" on Fedora installation guide.

Fixes [#373](https://github.com/kata-containers/runtime/issues/373).

Signed-off-by: Newton Jose <josephnewton001@gmail.com

### DIFF
--- a/install/fedora-installation-guide.md
+++ b/install/fedora-installation-guide.md
@@ -30,8 +30,9 @@ $ sudo -E VERSION_ID=$VERSION_ID dnf config-manager --add-repo \
 http://download.opensuse.org/repositories/home:/katacontainers:/release/Fedora\_$VERSION_ID/home:katacontainers:release.repo
 $ sudo -E dnf -y install kata-runtime kata-proxy kata-shim
 ```
+3. Check [hardware requirements](https://github.com/kata-containers/runtime/#hardware-requirements)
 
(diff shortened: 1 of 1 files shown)


jodh-intel:
Hi @n3wt0nSAN - thanks for raising this, but I suggest you make this change for all distros on your original #139 as the change to all the docs is related (so can all go on 1 commit).

PR #145, opened 2018-06-08
docs: Split docker install into separate docs

### LINKED ISSUE #144: split docker install steps into separate guides
The install guides currently assume all users want to use Docker. Since there are other container managers available, move the docker-specific commands into separate documents with links to them from the install guides.

jodh-intel:
The install guides assumed that users wanted to install Docker. Since there are other container managers, split the Docker-specific instructions into separate documents (with backlinks) and allow the user to choose between Docker or Kubernetes from the install guides.

Also, ensure the first mention of all the distro names in the install guides:

- Have the required asterisk after the name.
- Have a link to the website.

Fixes #144.

### DIFF
--- a/install/centos-installation-guide.md
+++ b/install/centos-installation-guide.md
@@ -8,53 +8,26 @@
 > - If you are installing on a system that already has Clear Containers or `runv` installed,
 >   first read [the upgrading document](../Upgrading.md).
 >
-> - This install guide is specific to integration with Docker.  If you want to use Kata
->   Containers through Kubernetes, see the "Developer Guide" [here](https://github.com/kata-containers/documentation/blob/master/Developer-Guide.md#if-you-want-to-run-kata-containers-with-kubernetes).
+> - If you do not want to copy or type all these instructions by hand, you can use the
+>   [`kata-manager`](https://github.com/kata-containers/tests/blob/master/cmd/kata-manager/kata-manager.sh)
+>   script to install the packaged system including your chosen container
+>   manager. Alternatively, you can generate a runnable shell script from
+>   individual documents using the
+>   [`kata-doc-to-script`](https://github.com/kata-containers/tests/blob/master/.ci/kata-doc-to-script.sh) script.
 
-This step is only required in case Docker is not installed on the system.
-1. Install the latest version of Docker with the following commands:
+1. Install the Kata Containers components with the following commands:
 
-```bash
-$ sudo yum -y install yum-utils
-$ sudo yum-config-manager --add-repo https://download.docker.com/linux/centos/docker-ce.repo
-$ sudo yum -y install docker-ce
-```
-For more information on installing Docker please refer to the
-[Docker Guide](https://docs.docker.com/engine/installation/linux/centos)
+   > **Note:** This installation channel is not secure since the repository currently
+   > redirects download URLs to `http`.
 
-2. Install the Kata Containers components with the following commands:
+   ```bash
+   $ source /etc/os-release
+   $ sudo yum -y install yum-utils
+   $ sudo -E VERSION_ID=$VERSION_ID yum-config-manager --add-repo "http://download.opensuse.org/repositories/home:/katacontainers:/release/CentOS_${VERSION_ID}/home:katacontainers:release.repo"
+   $ sudo -E yum -y install kata-runtime kata-proxy kata-shim
+   ```
 
-> **Note:** The repository redirects the download content to use `http`, be aware that this installation channel is not secure.
+2. Decide which container manager you will use and select the corresponding link that follows:
 
-```bash
-$ source /etc/os-release
-$ sudo -E VERSION_ID=$VERSION_ID yum-config-manager --add-repo \
-"http://download.opensuse.org/repositories/home:/katacontainers:/release/CentOS_${VERSION_ID}/home:katacontainers:release.repo"
-$ sudo -E yum -y install kata-runtime kata-proxy kata-shim
-```
-
-3. Configure Docker to use Kata Containers by default with the following commands:
-
-```bash
-$ sudo mkdir -p /etc/systemd/system/docker.service.d/
-$ cat <<EOF | sudo tee /etc/systemd/system/docker.service.d/kata-containers.conf
-[Service]
-ExecStart=
-ExecStart=/usr/bin/dockerd -D --add-runtime kata-runtime=/usr/bin/kata-runtime --default-runtime=kata-runtime
-EOF
-```
-
-4. Restart the Docker systemd service with the following commands:
-
-```bash
-$ sudo systemctl daemon-reload
-$ sudo systemctl restart docker
-```
-
-5. Run Kata Containers
-
-You are now ready to run Kata Containers:
-
-```
-$ sudo docker run -ti busybox sh
-```
+   - [Docker](docker/centos-docker-install.md)
+   - [Kubernetes](https://github.com/kata-containers/documentation/blob/master/Developer-Guide.md#if-you-want-to-run-kata-containers-with-kubernetes)
--- a/install/centos-installation-guide.md
+++ b/install/centos-installation-guide.md
@@ -2,7 +2,8 @@
 
 > **Notes:**
 >
-> - Kata Containers packages are available for CentOS version 7 (currently `x86_64` only).
+> - Kata Containers packages are available for [CentOS\*](https://www.centos.org)
+>   version 7 (currently `x86_64` only).
 >
 > - If you are installing on a system that already has Clear Containers or `runv` installed,
 >   first read [the upgrading document](../Upgrading.md).
--- a/install/docker/centos-docker-install.md
+++ b/install/docker/centos-docker-install.md
@@ -1,0 +1,54 @@
+# Install Docker for Kata Containers on CentOS
+
+> **Note:**
+>
+> - This guide assumes you have
+>   [already installed the Kata Containers packages](../centos-installation-guide.md).
+>
+> - If you do not want to copy or type all these instructions by hand, you can use the
+>   [`kata-manager`](https://github.com/kata-containers/tests/blob/master/cmd/kata-manager/kata-manager.sh)
+>   script to install the packaged system including your chosen container
+>   manager. Alternatively, you can generate a runnable shell script from
+>   individual documents using the
+>   [`kata-doc-to-script`](https://github.com/kata-containers/tests/blob/master/.ci/kata-doc-to-script.sh) script.
+
+1. Install the latest version of Docker with the following commands:
+
+   > **Note:** This step is only required if Docker is not installed on the system.
+
+   ```bash
+   $ sudo yum-config-manager --add-repo https://download.docker.com/linux/centos/docker-ce.repo
+   $ sudo yum -y install docker-ce
+   ```
+
+   For more information on installing Docker please refer to the
+   [Docker Guide](https://docs.docker.com/engine/installation/linux/centos).
+
+2. Configure Docker to use Kata Containers by default with the following commands:
+
+   ```bash
+   $ sudo mkdir -p /etc/systemd/system/docker.service.d/
+   $ cat <<EOF | sudo tee /etc/systemd/system/docker.service.d/kata-containers.conf
+   [Service]
+   ExecStart=
+   ExecStart=/usr/bin/dockerd -D --add-runtime kata-runtime=/usr/bin/kata-runtime --default-runtime=kata-runtime
+   EOF
+   ```
+
+3. Restart the Docker systemd service with the following commands:
+
+   ```bash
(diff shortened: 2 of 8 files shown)


jodh-intel:
Note: The CI should fail due to https://github.com/kata-containers/tests/issues/390.

jodh-intel:
Blocked on https://github.com/kata-containers/tests/pull/391.

jodh-intel:
Hi @klynnrif - ptal.

install/docker/ubuntu-docker-install.md
```diff
@@ -0,0 +1,48 @@
...
+   $ sudo -E apt-get -y install docker-ce
+   ```
+
+   For more information on installing Docker please refer to the
+   [Docker Guide](https://docs.docker.com/engine/installation/linux/ubuntu).
+
+2. Configure Docker to use Kata Containers by default with the following commands:
+
+   ```bash
+   $ sudo mkdir -p /etc/systemd/system/docker.service.d/
```
jcvenegas: Now we can also refactor this config in `install/docker/docker-kata-setup.md` :smile:

jodh-intel: Yes, I was thinking about that as this block is common for all the distros. It's difficult to get the right balance though as we don't want to send the user off having to read too many additional files.

wdyd @grahamwhaley, @egernst?

grahamwhaley: I agree we should keep the 'doc tree depth' shallow - and in this case, this is a command/sequence that changes infrequently, so I think it is probably best left replicated across the docs.
/me opens up an opportunity for @jodh-intel to mention that markdown is not the most extensive or flexible document format, and that I'm sure something like ReST has an `include` or similar fragment function we could use to do this better ;-)

install/centos-installation-guide.md
```diff
@@ -7,54 +7,20 @@
...
-5. Run Kata Containers
+   ```bash
+   $ source /etc/os-release
+   $ sudo yum -y install yum-utils
+   $ sudo -E VERSION_ID=$VERSION_ID yum-config-manager --add-repo "http://download.opensuse.org/repositories/home:/katacontainers:/release/CentOS_${VERSION_ID}/home:katacontainers:release.repo"
+   $ sudo -E yum -y install kata-runtime kata-proxy kata-shim
+   ```
 
-You are now ready to run Kata Containers:
+2. Decide which container manager you wish to use and click the corresponding link below:
```
grahamwhaley: I'm going to guess that @klynnrif might want `and click` to be rephrased. I do ;-) I might suggest s/click/follow/

install/docker/ubuntu-docker-install.md
jcvenegas: @jodh-intel and what about reference the users to the installation docs in docker and only document how to configure it to use kata as runtime. For would need to move the logic to install docker to kata-manager but we will have a more simple file.

install/centos-installation-guide.md
```diff
@@ -7,54 +7,20 @@
 >
 > - If you are installing on a system that already has Clear Containers or `runv` installed,
 >   first read [the upgrading document](../Upgrading.md).
->
-> - This install guide is specific to integration with Docker.  If you want to use Kata
->   Containers through Kubernetes, see the "Developer Guide" [here](https://github.com/kata-containers/documentation/blob/master/Developer-Guide.md#if-you-want-to-run-kata-containers-with-kubernetes).
-
-This step is only required in case Docker is not installed on the system.
```
grahamwhaley: (I know you didn't change this, but...) s/in case/ for either 'in the case that' or 'if' ?

install/docker/centos-docker-install.md
```diff
@@ -0,0 +1,44 @@
...
+   $ sudo systemctl daemon-reload
+   $ sudo systemctl restart docker
+   ```
+
+4. Run Kata Containers
+
+   You are now ready to run Kata Containers:
+
+   ```
+   $ sudo docker run -ti busybox sh
```
grahamwhaley: just a thought - I know we have this as a non-bash backtick as we cannot automatically test the `sh` invocation, as it would hang... how about we switch it to:

```bash
$ sudo docker run -ti busybox uname -a
```
(or the `bash -c 'uname -a'` if that is what is needed) - then we can both run that in the automation, and have it print out the in-VM kernel info to show we are in a VM and not on the host.

install/docker/ubuntu-docker-install.md
grahamwhaley: would that stop us automating it then though?

grahamwhaley: the testing that is...

install/docker/centos-docker-install.md
```diff
@@ -0,0 +1,44 @@
+# Installing Docker for Kata Containers on CentOS
+
+> **Note:**
+>
+> - This guide assumes you have
+>   [already installed the Kata Containers packages](../centos-installation-guide.md).
+
+1. Install the latest version of Docker with the following commands:
```
jcvenegas: I know  that we did not have issues recently with docker but do you think we need to ask to check versions.toml to verify the last known working/recommend  version.

install/docker/ubuntu-docker-install.md
jodh-intel: To summarise,

1) we have human-readable install docs which contain machine-parseable command blocks.
2) we have a script that can "execute" these docs.
3) we are currently automatically "testing" all changes to the install guides in the CI using (2).

I believe we also have two potentially conflicting user requirements:

a) some users want to install by running a script ("I have no time - just install it already!")
b) some user would rather review the commands we are asking them to run as `root` (understandably ;) and run those commands manually (assuming there aren't too many).

Although we could throw away the install docs and `kata-manager` and replace them with a big `install-kata.sh`, I don't think that's a good idea, nor would it be palatable to a lot of users (group (b) above).

I vote we retain (1)-(3) and:

- add some words to all the install docs explaining that if you are on a non-critical box, you *could* run `kata-manager.sh ...` to avoid typing in all the commands below, but we don't recommend it.
- add a big fat warning about doing this as it might be dangerous.
- add a new `--dry-run`-type option to `kata-manager.sh` that will generate the script for `install-*` commands, but won't execute it. That would allow users to review the script before executing it (as `root`).

grahamwhaley: I agree - ultimately the 'big fat script' action should be taken care of by distro packages - you install it through your distro package management thing, ideally from the distros repos, and it 'just works'(tm).

I think the `--dry-run` is going to be a useful alternative/addition - the docs can say at the top:
> **Note**: if you do not wish to copy or type all these instructions by hand, you can use the <blah blah> script to generate a runnable shell script from the - see <blah blah blah> for more details

install/centos-installation-guide.md
jodh-intel: Fixed.

install/centos-installation-guide.md
jodh-intel: Fixed.

install/docker/centos-docker-install.md
jodh-intel: like it - fixed.

install/docker/ubuntu-docker-install.md
jodh-intel: Hi @grahamwhaley - I've tweaked the wording a bit but now all the docs have the following standard wording as a Note at the top (ignore the bad formatting):

> - If you do not wish to copy or type all these instructions by hand, you can use either use the
>   [`kata-manager`](https://github.com/kata-containers/tests/blob/master/cmd/kata-manager/kata-manager.sh)
>   script to install the packaged system including your chosen container manager. Alternatively, you can generate
>   a runnable shell script from individual documents using the
>   [`kata-doc-to-script`](https://github.com/kata-containers/tests/blob/master/.ci/kata-doc-to-script.sh) script.

install/centos-installation-guide.md
```diff
@@ -2,58 +2,31 @@
...
-2. Install the Kata Containers components with the following commands:
+   ```bash
+   $ source /etc/os-release
+   $ sudo yum -y install yum-utils
+   $ sudo -E VERSION_ID=$VERSION_ID yum-config-manager --add-repo "http://download.opensuse.org/repositories/home:/katacontainers:/release/CentOS_${VERSION_ID}/home:katacontainers:release.repo"
+   $ sudo -E yum -y install kata-runtime kata-proxy kata-shim
+   ```
 
-> **Note:** The repository redirects the download content to use `http`, be aware that this installation channel is not secure.
+2. Decide which container manager you wish to use and follow the corresponding link below:
```
klynnrif: Suggested rewrite: 
2. Decide which container manager you will use and select the corresponding link that follows:

install/fedora-installation-guide.md
```diff
@@ -2,60 +2,31 @@
...
-[Docker Guide](https://docs.docker.com/engine/installation/linux/fedora)
+   ```bash
+   $ source /etc/os-release
+   $ sudo dnf -y install dnf-plugins-core
+   $ sudo -E VERSION_ID=$VERSION_ID dnf config-manager --add-repo http://download.opensuse.org/repositories/home:/katacontainers:/release/Fedora\_$VERSION_ID/home:katacontainers:release.repo
+   $ sudo -E dnf -y install kata-runtime kata-proxy kata-shim
+   ```
 
-2. Install the Kata Containers components with the following commands:
+2. Decide which container manager you wish to use and follow the corresponding link below:
```
klynnrif: Suggested rewrite:
2. Decide which container manager you will use and select the corresponding link that follows:

install/ubuntu-installation-guide.md
```diff
@@ -2,62 +2,31 @@
...
-> - Kata Containers packages are available for Ubuntu\* **16.04** and **17.10** (currently `x86_64` only).
+> - Kata Containers packages are available for [Ubuntu\*](https://www.ubuntu.com)
+>   versions **16.04** and **17.10** (currently `x86_64` only).
 >
 > - If you are installing on a system that already has Clear Containers or `runv` installed,
 >   first read [the upgrading document](../Upgrading.md).
 >
-> - This install guide is specific to integration with Docker.  If you want to use Kata
->   Containers through Kubernetes, see the "Developer Guide" [here](https://github.com/kata-containers/documentation/blob/master/Developer-Guide.md#if-you-want-to-run-kata-containers-with-kubernetes).
+> - If you do not wish to copy or type all these instructions by hand, you can use either use the
```
klynnrif: Lines 11-14 suggested rewrite:

If you do not want to copy or type all these instructions by hand, you can use the kata-manager script to install the packaged system including your chosen container manager. Alternatively, you can generate a runnable shell script from individual documents using the...

install/ubuntu-installation-guide.md
```diff
@@ -2,62 +2,31 @@
...
-[Docker Guide](https://docs.docker.com/engine/installation/linux/ubuntu)
+   ```bash
+   $ sudo sh -c "echo 'deb http://download.opensuse.org/repositories/home:/katacontainers:/release/xUbuntu_$(lsb_release -rs)/ /' > /etc/apt/sources.list.d/kata-containers.list"
+   $ curl -sL  http://download.opensuse.org/repositories/home:/katacontainers:/release/xUbuntu_$(lsb_release -rs)/Release.key | sudo apt-key add -
+   $ sudo -E apt-get update
+   $ sudo -E apt-get -y install kata-runtime kata-proxy kata-shim
+   ```
 
-2. Install the Kata Containers components with the following commands:
+2. Decide which container manager you wish to use and follow the corresponding link below:
```
klynnrif: Suggested rewrite:
2. Decide which container manager you will use and select the corresponding link that follows:

jodh-intel:

jodh-intel:
https://github.com/kata-containers/tests/pull/391 has been merged, so re-spinning the CI to hopefully see all these tests pass now...

install/docker/centos-docker-install.md
jodh-intel: That's a good question, but I'm not sure how we can add in that check without a bunch of code (as we'd have to remove the leading `v` from https://github.com/kata-containers/runtime/blob/master/versions.yaml#L130. But we'd also have the complication of documenting in the install guides how to check the version of docker that the distro-specific package manager is offering to install and then we've got to have shell logic to compare the semantic versions.

However, we *could* conceivably turn the problem round - add an extra check at the *end* of https://github.com/kata-containers/documentation/blob/master/.ci/test-install-docs.sh where we check the output of `docker version` after we've successfully created a container. If the `major.minor` part of the semver version string does not match `yq read versions.yaml externals.docker.version`, we could fail the test since it indicates `versions.yaml` needs to be updated.

I'm not sure that's a valid test in the sense that that new test would actually be testing `versions.yaml` (in a different repo).

wdyt @grahamwhaley, @chavafg?

Issue #158, opened 2018-06-14
 developer-guide: Use correct kernel config file name


nitkon:
Currently, the developer-guide uses amd64 instead of x86_64 as the prefix when pulling the kernel config file. Fix this issue.

PR #159, opened 2018-06-14
 developer-guide: use the correct kernel config file name

nitkon:
kernel_arch was being set to amd64 instead of x86_64
on intel. The kernel config file name starts with
x86_64 and hence this needs to be fixed.

Fixes:  #158

Signed-off-by: Nitesh Konkar <niteshkonkar@in.ibm.com>


jodh-intel:
lgtm

PR #164, opened 2018-06-14
docs: Add relevant URLs and fix grammar

nitkon:
Fixes:  #163

Signed-off-by: Nitesh Konkar <niteshkonkar@in.ibm.com>

### DIFF
--- a/architecture.md
+++ b/architecture.md
@@ -47,8 +47,8 @@ and `kata-runtime`.
 
 ![Docker and Kata Containers](arch-images/docker-kata.png)
 
-`kata-runtime` creates a QEMU\*/KVM virtual machine for each container or pod
-the Docker engine or Kubernetes' `kubelet` creates.
+`kata-runtime` creates a QEMU\*/KVM virtual machine for each container or pod,
+the Docker engine or Kubernetes' `kubelet` creates respectively.
 
 The container process is then spawned by
 [agent](https://github.com/kata-containers/agent), an agent process running
@@ -102,7 +102,7 @@ for example `pc` and `q35` on x86 systems and `virt` on ARM systems. Kata Contai
 default machine type is `pc`. The default machine type and its [`Machine accelerators`](#Machine-accelerators) can
 be changed by editing the runtime [`configuration`](#Configuration) file.
 
-The follow QEMU features are used in Kata Containers to manage resource constraints, improve
+The following QEMU features are used in Kata Containers to manage resource constraints, improve
 boot time and reduce memory footprint:
 
 - Machine accelerators.
@@ -160,7 +160,7 @@ Kata Containers supports both an `initrd` and `rootfs` based minimal guest image
 
 ##### Root filesystem image
 
-The default root filesystem image, sometimes referred to as the "mini O/S", is a
+The default packaged root filesystem image, sometimes referred to as the "mini O/S", is a
 highly optimized container bootstrap system based on [Clear Linux](https://clearlinux.org/). It provides an extremely minimal environment and
 has a highly optimized boot path.
 
@@ -230,8 +230,7 @@ Most users will not need to modify the configuration file.
 The file is well commented and provides a few "knobs" that can be used to modify
 the behavior of the runtime.
 
-The configuration file is also used to enable runtime debug output (see
-some url to documentation on how to enable debug).
+The configuration file is also used to enable runtime [debug output](https://github.com/kata-containers/documentation/blob/master/Developer-Guide.md#enable-full-debug).
 
 ### Significant OCI commands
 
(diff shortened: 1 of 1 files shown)


jodh-intel:
Thanks @nitkon!

lgtm

@klynnrif - ptal.

architecture.md
```diff
@@ -230,8 +230,7 @@ Most users will not need to modify the configuration file.
 The file is well commented and provides a few "knobs" that can be used to modify
 the behavior of the runtime.
 
-The configuration file is also used to enable runtime debug output (see
-some url to documentation on how to enable debug).
```
grahamwhaley: Nice - can't believe we left 'some url' in there!

architecture.md
```diff
@@ -361,7 +359,7 @@ process representing this container process.
 Communication with the VM can be achieved by either `virtio-serial` or, if the host
 kernel is newer than v4.8, a virtual socket, `vsock` can be used. The default is `virtio-serial`.
 
-The VM will likely be running multiple container processes.  In the event `virtio-serial`
+The VM likely will be running multiple container processes.  In the event `virtio-serial`
```
grahamwhaley: Heh, my head reads it better as the original way around - I think this is one for @klynnrif :-)

architecture.md
nitkon: @grahamwhaley : `Call into the post-start hooks. Usually, this is a no-op since nothing is provided (this needs clarification)` [here](https://github.com/kata-containers/documentation/blob/master/architecture.md#start) also needs to be removed after providing clarification?

grahamwhaley: Ah, indeed @nitkon - we should try to clarify those.
@sboeuf - are you able to add any clarification to these two places in the arch doc:

https://github.com/kata-containers/documentation/blame/master/architecture.md#L278
```
2. Call into the post-start hooks. Usually, this is a no-op since nothing is provided
(this needs clarification)
```

https://github.com/kata-containers/documentation/blame/master/architecture.md#L347
```
1. Ask the container status by checking information stored on disk. (clarification needed)
```

If we can provide clarification and wording, maybe @nitkon would be kind enough to add the fixes to this PR :-)

grahamwhaley:
Looking at https://github.com/kata-containers/osbuilder/tree/master/rootfs-builder, yes, I believe that is true @nitkon :-)

jodh-intel:
Hi @nitkon.

> The default root filesystem image, sometimes referred to as the "mini O/S", is a highly optimized container bootstrap system based on Clear Linux.

That is still 100% correct in that the **default packaged rootfs image** is indeed based on Clear Linux:

- https://github.com/kata-containers/packaging/blob/master/kata-containers-image/build_image.sh#L23

Whereas the default packaged initrd image is based on alpine:

- https://github.com/kata-containers/packaging/blob/master/kata-containers-image/build_image.sh#L28

But the default "image type" ("image" or "initrd") is "image", hence a Clear Linux rootfs image:

- https://github.com/kata-containers/packaging/blob/master/runtime/debian.rules-template#L43
- https://github.com/kata-containers/packaging/blob/master/runtime/kata-runtime.spec-template#L95

But feel free to elaborate on the topic to make it clearer if you wish.

This PR is sitting waiting to land (green button), so please let us know whether it should be merged or whether you plan to add further changes to it.

nitkon:
@jodh-intel : Nice! Thanks for your detailed response. I think we should document **default packaged rootfs image is Clear Linux** whereas the **default packaged initrd image is based on alpine** and the **default "image type" ("image" or "initrd") is "image", hence a Clear Linux rootfs image**

I sincerely believe we should improve the documentation for *packaging* repo as its README is not enough.
(Can these details be added in the README.me?) 

Also, I am waiting for @sboeuf to reply on the "ask" in the previous comment by @grahamwhaley. If he has nothing to add we can merge the PR else I will update it based on his inputs

jodh-intel:
@nitkon - We welcome PRs... :) If not, please just raise an issue so it isn't forgotten about.

nitkon:
Cool. Then we are good to merge.

architecture.md
```diff
@@ -331,7 +329,7 @@ cannot be deleted unless the OCI runtime is explicitly being asked to, by using
 
 If the sandbox is not stopped, but the particular container process returned on
 its own already, the `kata-runtime` will first go through most of the steps a `kill`
-would go through for a termination signal. After this (or simply this if the sandboxIDwas already stopped), then `kata-runtime` will: If the sandbox was already stoppedfollowed by:
+would go through for a termination signal. After this (or simply this if the sandboxID was already stopped), then `kata-runtime` will: If the sandbox was already stopped followed by:
```
klynnrif: Including "if the sandbox was already stopped" twice in this sentence seems redundant, but I'm concerned I will change the meaning here :) I'll suggest the following rewrite: 
After this process, or if the sandboxID was already stopped to begin with, `kata-runtime` will:

jodh-intel: @nitkon - this looks like the last block on this PR.

jodh-intel:
Marking dnm until @klynnrif approves.

architecture.md
nitkon: @jodh-intel @klynnrif : Updated my PR. Please take a look.

PR #191, opened 2018-07-05
Document compatibility of component versions

### LINKED ISSUE #177: Document that all component versions must match for current compatibility
We need some words in the following documents to state that if you are using runtime release X, all other components must (currently) also be at version X:

- https://github.com/kata-containers/documentation/blob/master/Releases.md#versioning
- https://github.com/kata-containers/documentation/blob/master/Upgrading.md
  (pointer to info in the Releases.md)

jodh-intel:
Add new "Component version compatibility" section to the Releases document that explains all component versions must match.

Also, update the developer guide and upgrading document to refer to the "Component version compatibility" section of the Releases doc.

Fixes #177.

### DIFF
--- a/Developer-Guide.md
+++ b/Developer-Guide.md
@@ -2,6 +2,7 @@
 * [Assumptions](#assumptions)
 * [Initial setup](#initial-setup)
 * [Requirements to build individual components](#requirements-to-build-individual-components)
+* [Individual component compatibility](#individual-component-compatibility)
 * [Build and install the Kata Containers runtime](#build-and-install-the-kata-containers-runtime)
     * [Check hardware requirements](#check-hardware-requirements)
     * [Configure to use initrd or rootfs image](#configure-to-use-initrd-or-rootfs-image)
@@ -74,6 +75,22 @@ You need to install the following to build Kata Containers components:
 
 - `make`.
 - `gcc` (required for building the shim and runtime).
+
+# Individual component compatibility
+
+Using versions of individual components newer than the last official released
+version might result in
+[compatibility issues](Releases.md#component-version-compatibility).
+
+If you have only updated a subset of the Kata components and notice strange
+behavior, we recommend you update the remaining components to ensure you are
+using the latest versions of all of them. Also, remember to
+[create a new image containing the latest agent version](#create-and-install-rootfs-and-initrd-image).
+
+If you still have compatibility issues after following the previous
+instructions, refer to the
+[troubleshooting section](#troubleshoot-kata-containers) and
+[get in contact with us](https://github.com/kata-containers/community/blob/master/CONTRIBUTING.md#contact).
 
 # Build and install the Kata Containers runtime
 
--- a/Developer-Guide.md
+++ b/Developer-Guide.md
@@ -1,3 +1,5 @@
+# Kata Containers Developer Guide
+
 * [Warning](#warning)
 * [Assumptions](#assumptions)
 * [Initial setup](#initial-setup)
--- a/Releases.md
+++ b/Releases.md
@@ -37,7 +37,10 @@ Semantic versioning is used since the version number is able to convey clear inf
 
 ## Components
 
-A new release will result in all Kata components being given a new [version](#versioning), even if no changes were made to that component since the last version. The version for a release is **identical** for all  components.
+A new release results in all Kata components being given a new
+[version](#versioning), even if there are no changes to that component since
+the last version. The version for a release is **identical** for all
+components.
 
 This strategy allows diagnostic tools such as `kata-runtime kata-env` to record full version details of all components to help with problem determination.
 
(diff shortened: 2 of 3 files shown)


jodh-intel:
Hi @klynnrif, @intelkevinputnam, @grahamwhaley - ptal.

Releases.md
```diff
@@ -42,6 +43,16 @@ This strategy allows diagnostic tools such as `kata-runtime kata-env` to record
 
 Note that although hypervisor and guest kernel both have versions, these are not updated for new releases as they are not core components developed by the Kata community.
 
+### Component version compatibility
+
+Semantic versioning allows users to determine how a new version of a
+particular component relates to the previous version of that component.
+However, it is not possible to compare two *different* Kata component
```
grahamwhaley: Can you explain or expand on why it is 'not possible to compare...' - it's not quite clear to me here what you are trying to explain.

Developer-Guide.md
```diff
@@ -75,6 +76,22 @@ You need to install the following to build Kata Containers components:
 - `make`.
 - `gcc` (required for building the shim and runtime).
 
+# Individual component compatibility
+
+Using newer versions of individual components than the last offical released
```
grahamwhaley: potential reword, and a typo s/offical/official/

`Using versions of individual components newer than the last official released`

Developer-Guide.md
```diff
@@ -75,6 +78,22 @@ You need to install the following to build Kata Containers components:
...
+version may result in
```
klynnrif: may => might: 
version might result in

Developer-Guide.md
```diff
@@ -75,6 +78,22 @@ You need to install the following to build Kata Containers components:
...
+[compatibility issues](Releases.md#component-version-compatibility).
+
+If you have only updated a subset of the Kata components and notice strange
+behaviour, we recommend you update the remaining components to ensure you are
```
klynnrif: behaviour => behavior

Developer-Guide.md
```diff
@@ -75,6 +78,22 @@ You need to install the following to build Kata Containers components:
...
+using the latest versions of all of them. Remember to also [create a new image
```
klynnrif: ... Also, remember to [create a new image

Developer-Guide.md
```diff
@@ -75,6 +78,22 @@ You need to install the following to build Kata Containers components:
...
+containing the latest agent
+version](#create-and-install-rootfs-and-initrd-image).
+
+If this still does not work, refer to the
```
klynnrif: Suggested rewrite: If you still have compatibility issues after following the previous instructions, refer to the

Releases.md
```diff
@@ -34,6 +35,24 @@ Semantic versioning is used since the version number is able to convey clear inf
 
   A major release will also likely require a change of the container manager version used, for example Docker\*. Please refer to the release notes for further details.
 
+## Components
+
+A new release will result in all Kata components being given a new [version](#versioning), even if no changes were made to that component since the last version. The version for a release is **identical** for all  components.
```
klynnrif: Suggested rewrite: 
A new release results in all Kata components being given a new [version](#versioning), even if there are no changes to that component since the last version.

jodh-intel:

Releases.md
jodh-intel: Nor me! :smile:

I think we're all struggling to really understand what this actually means, but I'm trying to find words, so if you can suggest some based on what follows, I'm all ears! ;)

It's rather difficult to write this. What I'm trying to state is that although we "use" semver for each component, since we require all components to have the same version to be compatible, we require our own... cough.. interpretation of what semver means. It could be argued that semver is implicitly only meant for comparing versions *of the same component*.

However since we're imposing the extra requirement that all components need to be upgraded in lock-step (all components having the same version), a gRPC protocol change (what would nominally be an "API breakage" for a single component) for example can be handled by a MINOR version number change only (since this detail is invisible to the user *assuming they upgrade all components in lock-step*).

For example, if the proxy is at version `1.2.0` but the other components are all at `1.1.0`, there are no guarantees the system will work.

I'll also probably add words based on @gnawux's comments to the ML (http://lists.katacontainers.io/pipermail/kata-dev/2018-June/000232.html) too.

egernst: Sure, but 1.1.1 and 1.1.0 should be fine.  Right?

jodh-intel: @grahamwhaley, @egernst - should we just drop the "Component version compatibility" section entirely then do you think?

jodh-intel: Re-ping @grahamwhaley, @egernst.

jodh-intel: Thoughts guys...? :)

grahamwhaley: Hi @jodh-intel - I'm prone to put this PR on hold, pretty much just due to this semver related section, until we have the Kata versioning and release flow defined - and then we can reference that.
I believe @jcvenegas is planning a first draft around releases, versioning and backport/lifetimes - let's wait for that.
As such, I'm going to mark this PR as WIP, just to make it clear it is pending an update.

jodh-intel:
Closing as this is covered by https://github.com/kata-containers/documentation/blob/master/Stable-Branch-Strategy.md.

PR #194, opened 2018-07-13
Add kata runtime to Dockers daemon.json

morph027:
One can also add runtimes in daemon.json without overriding systemd stuff.

### DIFF
--- a/install/docker/ubuntu-docker-install.md
+++ b/install/docker/ubuntu-docker-install.md
@@ -28,7 +28,9 @@
    For more information on installing Docker please refer to the
    [Docker Guide](https://docs.docker.com/engine/installation/linux/ubuntu).
 
-2. Configure Docker to use Kata Containers by default with the following commands:
+2. Configure Docker to use Kata Containers by default with one of the following methods:
+
+2.1 Using systemd
 
    ```bash
    $ sudo mkdir -p /etc/systemd/system/docker.service.d/
@@ -37,6 +39,18 @@
    ExecStart=
    ExecStart=/usr/bin/dockerd -D --add-runtime kata-runtime=/usr/bin/kata-runtime --default-runtime=kata-runtime
    EOF
+   ```
+
+2.2 Using Docker daemon.json
+
+   Add the following to your `/etc/docker/daemon.json`
+
+   ```
+   "runtimes": {
+    "kata-runtime": {
+      "path": "/usr/bin/kata-runtime"
+    }
+   }
    ```
 
 3. Restart the Docker systemd service with the following commands:


grahamwhaley:
Nice addition @morph027 :-) Yeah, I think I stumbled upon this the other day - this could be really useful.

A couple of things to fix up/consider on this PR:
- The CIs are failing, I suspect because you need to modify the PR/commits to match the [patch format guidelines](https://github.com/kata-containers/community/blob/master/CONTRIBUTING.md#patch-format) for the project. Including at least:
  - having a 'subject: msg' format for the subject line
  - having a 'Fixes: #nnn' in at least one commit
-  The numbered list indentation looks broken on your doc (you can view it in your branch [here](https://github.com/morph027/documentation/blob/d09af6161bbc7a3bedc0f4afcbc37a70720d27ed/install/docker/ubuntu-docker-install.md) for instance). I *think* you can fix that with some indentation in the markdown, but I'm not sure - I'll try a little experiment below...
- I don't suppose you'd like to fix the other three files in the [same directory](https://github.com/kata-containers/documentation/tree/master/install/docker) would you?, presuming they have the same text (which, tbh, is some duplication we could do with removing, but it would make our docs tree/click-depth a bit cumbersome) ? :-)

OK, let's try an indented numbered list. OK, best I can do quickly is to make a sub-list indented with 4 spaces and starting at `1)` again, so this body:

```
1) one
2) two
    1) two-dot-two
3) three
```

Makes this output:

1) one
2) two
    1) two-dot-two
3) three

morph027:
Thanks for the hints...i'll create a new one.

grahamwhaley:
:-) - np - you didn't have to make a new one though - you could just fix up your branch and then 'force push' that with git (`git push -f <origin> <branch>`), and github would DTRT - that is how we normally work - update the same branch/PR as we get feedback.

install/docker/ubuntu-docker-install.md
```diff
@@ -39,6 +41,18 @@
...
    ```
 
+2.2 Using Docker daemon.json
+
+   Add the following to your `/etc/docker/daemon.json`
+
+   ```
+   "runtimes": {
+    "kata-runtime": {
+      "path": "/usr/bin/kata-runtime"
```
amshinde: @morph027 Could you also add default runtime here `"default-runtime": "kata"` ?

install/docker/ubuntu-docker-install.md
```diff
@@ -28,7 +28,9 @@
    For more information on installing Docker please refer to the
    [Docker Guide](https://docs.docker.com/engine/installation/linux/ubuntu).
 
-2. Configure Docker to use Kata Containers by default with the following commands:
+2. Configure Docker to use Kata Containers by default with one of the following methods:
+
+2.1 Using systemd
```
amshinde: Formatting seems to be off here.

grahamwhaley:
both of those fixed on the replacement PR #195 I believe :-)

amshinde:
ohh :) Will take a look at the other PR.

PR #196, opened 2018-07-23
WIP: gRPC Best practices.

jon:


best-practices/grpc-apis.md
```diff
@@ -0,0 +1,105 @@
...
+
+## Guidelines
+
+### GL1. Have Clearly Defined Compatibility Windows
+
+TODO: Change this guideline to encompass Kata's agreed-upon compatibiltiy
+policy.
+
+It is critical to have a clear policy on what upgrade and downgrade paths are
+supported. One policy that has been generally successful for other large
```
jodh-intel: We're going to need a careful testing matrix for this too. And these compatibility tests are going to have to block a release.

/cc @chavafg.

best-practices/grpc-apis.md
```diff
@@ -0,0 +1,105 @@
...
+projects (Google Compute Engine, in particular) it to support unbounded forward
+upgrades and single-release downgrades.
+
+### GL2. The Scope of Compatibility is Only Releases that may Mix
+
+Many compatibility problems can been addressed be avoiding mixing components.
```
jodh-intel: s/be avoiding mixing components/by avoiding mixing component versions/ ?

best-practices/grpc-apis.md
```diff
@@ -0,0 +1,105 @@
...
+For example, if each Kata release includes both the agent and the runtime, the
+agent can be
```
jodh-intel: Incomplete sentence :)

best-practices/grpc-apis.md
```diff
@@ -0,0 +1,105 @@
...
+
+### GL3. Redundancy is a Useful Tool
+
+While it can be aesthetically displeasing, redundantly encoding similar (or
+identical) information can allow slow migration from one behavior
```
jodh-intel: And another. I'm rather intrigued to read the rest of this sentence though :)

best-practices/grpc-apis.md
```diff
@@ -0,0 +1,105 @@
...
+
+### GL4. Changes to Defaults Should Straddle Compatibility Boundaries
+
+As per R2 above, new fields cannot generally be introduced and used in the same
+release. More broadly, new behaviors (typically new features) should not become
+the default until the _oldest_ release not supporting that behavior is no longer
+in the downgrade compatibiltiy window.
```
jodh-intel: Typo.

best-practices/grpc-apis.md
intelkevinputnam: behavior _to another._

best-practices/grpc-apis.md
intelkevinputnam: can be ?????

best-practices/grpc-apis.md
intelkevinputnam: can _be_ addressed

best-practices/grpc-apis.md
```diff
@@ -0,0 +1,105 @@
...
+
+### R4. Rarely Change Field Types
+
+Related to the R1, while not always strictly unsafe it is usually unwise to
+change the type of an existing field. In most cases it is better to inroduce a
+new field with the new type and slowly phase out the old field once it has left
+the backward compatibility window.
+
+Exceptions to this rule can include converting enum types to int32 or int64,
+provided no values other than the original enum definition are used
```
intelkevinputnam: are used.

best-practices/grpc-apis.md
```diff
@@ -0,0 +1,105 @@
...
+the strictest sense this means that new features cannot be "enabled" until one
+release after they are deemed "working". As a concrete example, when Google
+Compute Engine introduced multi-queue networking we did so one release later
+than when it was "fully supported" in GCE production. This allowed us to safely
+roll-back the release when a bug elsewhere was encountered while not bricking
+VMs that had booted after multi-queue had been enabled.
+
+TODO: Find a concrete example from Kata rather than from an external source.
+
+### R3. Enums are Problematic
```
intelkevinputnam: This needs to be made into a rule: "Do Not Use Enums Unless There Is No Other Alternative," or similar.

best-practices/grpc-apis.md
```diff
@@ -0,0 +1,105 @@
...
+```
+
+### R2. Avoid Introducing and Enable a Feature in the same Release
+
+Maintaining downgrade compatibility when launching a new release requires that
+the release not include any features not supported by the previous release. In
+the strictest sense this means that new features cannot be "enabled" until one
+release after they are deemed "working". As a concrete example, when Google
+Compute Engine introduced multi-queue networking we did so one release later
+than when it was "fully supported" in GCE production. This allowed us to safely
```
intelkevinputnam: release _after_ it was "fully...

best-practices/grpc-apis.md
```diff
@@ -0,0 +1,105 @@
...
+
+```proto
+reserved 1, 3;
+
+string name = 2;
+
+int32 size = 4;  // <-- The first available field number is 4.
+```
+
+### R2. Avoid Introducing and Enable a Feature in the same Release
```
intelkevinputnam: _Enabling_ a Feature in the _Same_ Release

best-practices/grpc-apis.md
```diff
@@ -0,0 +1,105 @@
...
+
+## Rules
+
+### R1. Never, _EVER_ Re-use Protcol Buffer Field Numbers
+
+Every protocol buffer field has an identifying number. When
+[encoded](https://developers.google.com/protocol-buffers/docs/encoding) on the
+wire, this number and the field's type are the only information that is actually
+passed. The result is that it is _never_ safe to re-use a field number. Protocol
+buffers provides the `reserved` keyword allowing field numbers to be marked as
```
intelkevinputnam: buffers _provide_ the ...

best-practices/grpc-apis.md
```diff
@@ -0,0 +1,105 @@
...
+The most important takeaway from this document should be that changing protocol
+buffers used across software releases requires careful planning and analysis of
+behavior both in future releases and past releases. Every change to a protobuf
+defintion should include an analysis of how future releases will cope with the
+field if it is not present as well as how past releases will behave given that
+they will be blind to the new field.
+
+## Rules
+
+### R1. Never, _EVER_ Re-use Protcol Buffer Field Numbers
```
raravena80: Typo => Protocol

best-practices/grpc-apis.md
```diff
@@ -0,0 +1,105 @@
...
+string name = 2;
+
+int32 size = 4;  // <-- The first available field number is 4.
+```
+
+### R2. Avoid Introducing and Enable a Feature in the same Release
+
+Maintaining downgrade compatibility when launching a new release requires that
+the release not include any features not supported by the previous release. In
+the strictest sense this means that new features cannot be "enabled" until one
```
raravena80: strictest sense, <- (comma)

best-practices/grpc-apis.md
```diff
@@ -0,0 +1,105 @@
...
+than when it was "fully supported" in GCE production. This allowed us to safely
+roll-back the release when a bug elsewhere was encountered while not bricking
+VMs that had booted after multi-queue had been enabled.
+
+TODO: Find a concrete example from Kata rather than from an external source.
+
+### R3. Enums are Problematic
+
+Related to R2, enum types are problematic as adding additional values in a new
+releases can create undefined behavior in prior releases.
```
raravena80: typo: in a new release

best-practices/grpc-apis.md
WeiZhang555: Kata agent works as grpc server, kata-runtime works as grpc client, so I think an example matrix can be:


  | Runtime v1 | Runtime v2 | Runtime v3 | Runtime v4
-- | -- | -- | -- | --
Agent v1 | √ | √ | √ | √
Agent v2 | √ | √ | √ | √
Agent v3 | × | √ | √ | √
Agent v4 | × | × | √ | √

Do I understand it right? @jon

best-practices/grpc-apis.md
WeiZhang555: It must be stripped when copy :smile:

WeiZhang555:
Related to: 
1. [Version the gRPC protocol ](https://github.com/kata-containers/agent/issues/272)
2. [Kata Components should support "Live-Upgrade“ ](https://github.com/kata-containers/runtime/issues/492)

best-practices/grpc-apis.md
```diff
@@ -0,0 +1,105 @@
...
+the default until the _oldest_ release not supporting that behavior is no longer
+in the downgrade compatibiltiy window.
+
+### GL5. On-the-Wire Versioning is Rarely Useful
+
+Very few protocol buffers at Google include version information as part of the
+wire format (either explicitly or implicitly). For example, the Google Compute
+Engine API is still on "version 1" despite dozens of calls and arguments being
+added since its inception. This is possible because new features are rolled out
+as "additions" to existing APIs, and by following R2 above.
```
WeiZhang555: As kata-agent will keeps running after kata-runtime is upgraded, I think we need a mechanism to let   kata-runtime know which field isn't supported by this version of kata-agent, and kata-runtime should downgrade it's grpc message version to appropriate version for better adapt to the agent.  so question is, without on wire versioning support, how could kata-runtime do this？

WeiZhang555: One way I think is, we give the on-disk storage config of a POD a version number, and this version number will represent the kata-agent version, kata-runtime will send appropriate message according to the on-disk version(which equals to agent version). What do you think, do you have better practice to share?

best-practices/grpc-apis.md
jodh-intel: +1. We could have an automated check to forbid enums too of course. Thoughts @sboeuf, @bergwolf?

best-practices/grpc-apis.md
jodh-intel: I think that @jon is suggesting we should not perform any version checks.

A possible solution which was mentioned briefly yesterday is that we could introduce a new `RestartAgent` gRPC service. When we can guarantee that all agent versions support that service (and assuming that adding new services to the protocol will not stop `RestartAgent` from working), when the runtime is upgraded, it can call `RestartService` for running pods. That will cause the agent to re-exec itself, reloading a newer binary version (which supports the newer gRPC protocol version). Once done, the runtime is guaranteed that it can use the full (new) gRPC protocol version when talking to the agent. This would potentially require the agent to pass state between the original instance and the newly-exec'd instance but that should be achievable.

Alternatively, rather than re-execing the agent, we could apparently squirt bytes down the wire via gRPC from the runtime to the agent so that the agent could read the byte stream and create a new agent instance out of it (write the bytes to disk and re-exec?). I think @jon mentioned that would be slower though?

I'd vote for the re-exec option as it's proven and used by init systems such as systemd and upstart.

One issue with both approaches is being able to rewrite the agent binary. This might not be possible or desirable. We could share the agent into all pods images via 9p and then the runtime would only need to update the (host) agent binary file and ask all the running agent instances to re-exec to use it. Again, performance might take a hit here due to 9p.

Another concern is of course security - arbitrarily rewriting files isn't ideal. We'd need to think carefully about checksums, security tokens, gpg signatures, etc.

grahamwhaley: How long it takes to get a new agent binary into the VM, and how, I'm not so worried about. This will be a very rare sequence (it's not like something we would be doing every minute, or even daily - this is probably a once every 3 month or more type scenario I would think).
How we hand over the state from the old agent to the new I think will be an interesting challenge - but, as you say, should be achievable.

jodh-intel: For Upstart, we used a pipe - see https://wiki.ubuntu.com/FoundationsTeam/Specs/QuantalUpstartStatefulReexec#State-Passing for the logic.

best-practices/grpc-apis.md
```diff
@@ -0,0 +1,105 @@
+# Evolving gRPC APIs &mdash; Best Practices
```
grahamwhaley: Most of our md docs have a toc at the top or just below the first level title - consider if it is worth adding one to this doc as well.

best-practices/grpc-apis.md
```diff
@@ -0,0 +1,105 @@
...
+int32 size = 4;  // <-- The first available field number is 4.
+```
+
+### R2. Avoid Introducing and Enable a Feature in the same Release
+
+Maintaining downgrade compatibility when launching a new release requires that
+the release not include any features not supported by the previous release. In
+the strictest sense this means that new features cannot be "enabled" until one
+release after they are deemed "working". As a concrete example, when Google
+Compute Engine introduced multi-queue networking we did so one release later
```
grahamwhaley: (tm) ?

best-practices/grpc-apis.md
```diff
@@ -0,0 +1,105 @@
...
+than when it was "fully supported" in GCE production. This allowed us to safely
+roll-back the release when a bug elsewhere was encountered while not bricking
+VMs that had booted after multi-queue had been enabled.
+
+TODO: Find a concrete example from Kata rather than from an external source.
```
grahamwhaley: In case it helps - I think there is some non-backwards compat change in one of the components between the current [head](https://github.com/kata-containers/runtime/commit/67b5841153424582db620a96eeaf8eb365e1c12e), and the v1.1.0 tags. If I roll the runtime back to 1.1.0 for instance I can not launch docker containers. I've not figured out where the break was though - if anybody knows (and I suspect an agent/proxy relationship maybe), please speak up...

best-practices/grpc-apis.md
grahamwhaley: I've not looked at gRPC enums (and I'm presuming this is only relating to gRPC enums) - but is there a certain class of enums we are talking about here. If one had a linear/sequential set of enums that only ever grew on the end for instance, is that problem free?

best-practices/grpc-apis.md
```diff
@@ -0,0 +1,105 @@
...
+
+### R3. Enums are Problematic
+
+Related to R2, enum types are problematic as adding additional values in a new
+releases can create undefined behavior in prior releases.
+
+### R4. Rarely Change Field Types
+
+Related to the R1, while not always strictly unsafe it is usually unwise to
+change the type of an existing field. In most cases it is better to inroduce a
```
grahamwhaley: s/inroduce/introduce/

jodh-intel:
Hi @jon,

Please could you update your commit as if you look at the Travis log:

```
Found 1 commit between commit 71a506c739f69ea6a431ba9f3bfe6e2bff7816f3 and branch master
ERROR: Commit 71a506c739f69ea6a431ba9f3bfe6e2bff7816f3: pure whitespace body
commits for the project.
```

raravena80:
@jon ping, any updates? Thx!

jodh-intel:
Hi @jon - could you tal at this when you get a chance? :smile:

jon:
Yeah,  like that plan -- I'll take a stab at a quick set of edits to this
over the next few days to cover the feedback and the issues that I recall
coming up in Arch Council meetings. Perhaps future addenda can be
crowdsourced from there :)

Jon

raravena80:
@jon ping from the kata herder :)

sboeuf:
@jon what's the status here?

jodh-intel:
Yep - I'm afraid this PR has the dubious honour of being the second oldest (behind the vintage https://github.com/kata-containers/proxy/pull/74).

If anyone has cycles, we could invoke the [assisted process](https://github.com/kata-containers/community/blob/master/CONTRIBUTING.md#assisted-pr-workflow) on this one?

raravena80:
@jon ping (from your weekly Kata herder)

jodh-intel:
Hi @jon - this is your friendly periodic ping for this ol' :chestnut: (<-- `:chestnut:` :smile:)

raravena80:
@jon any chance to take a look at this?

jodh-intel:
:cricket:'s :smile:

jodh-intel:
Nice! Slack shows this emoji as I intended it! :)

jodh-intel:
Unless we have any volunteers to update this using the [assisted workflow](https://github.com/kata-containers/community/blob/master/CONTRIBUTING.md#assisted-pr-workflow), I propose we close this.

grahamwhaley:
let's hear from @jon - is this good to merge as is (better than nothing), has it rotted, or does it need a mild refresh. Guide us @jon on what to do with this pls :-)

raravena80:
@jon any updates? 😄

grahamwhaley:
@jon - heeelllooowweeee.... :-) @kata-containers/architecture-committee - can we make a decision, to either:
- have some volunteer(s) take this through assisted cleanup (pretty much probably apply the feedback as is and merge)
- or maybe @jon can fix this??
- or, can/shall we close this down?

jodh-intel:
Hi @jon and @kata-containers/architecture-committee. This PR has now been open for 291 days. Since the requested feedback still hasn't been applied and since nobody has come forward to handle this (I think you are in fact the only one who could do this given your knowledge of gRPC), I'm going to close it.

Of course "close" does not mean "delete", so it can be reopened at any time if anyone has spare cycles at some point.

PR #203, opened 2018-07-28
Fix debug console socket docs

grahamwhaley:
Looks like the debug console socket location has moved. Update it.

Also minor grammatical fixups.


jodh-intel:
Trivial change - merging...

PR #213, opened 2018-08-14
docs: Change install URLs to https

### LINKED ISSUE #212: Update install docs for https urls
All the installation guides that specify OBS URLS...

- https://github.com/kata-containers/documentation/tree/master/install
- https://github.com/kata-containers/documentation/tree/master/install/docker

... currently specify **http** URLs along with the following warning:

> **Note:** This installation channel is not secure since the repository currently redirects download URLs to `http`.

This is now very close to being fully resolved as download URLs for Ubuntu, Fedora, RHEL and CentOS all correctly redirect from https -> https fwics:

However, we're still waiting for OpenSuSE and SLES to be updated - see https://github.com/kata-containers/documentation/pull/83#issuecomment-412438187.

Once this is done, we can do the following to all the install docs:

- Update the OBS URLs from http to https.
- Remove the warning.

/cc @adrianschroeter, @vrothberg.

jodh-intel:
Now that the OBS server has enabled full https (no more redirecting to
http for the actual package downloads), remove the warning and switch to
specifying an https repository URL.

Fixes #212.

### DIFF
--- a/install/centos-installation-guide.md
+++ b/install/centos-installation-guide.md
@@ -17,13 +17,10 @@
 
 1. Install the Kata Containers components with the following commands:
 
-   > **Note:** This installation channel is not secure since the repository currently
-   > redirects download URLs to `http`.
-
    ```bash
    $ source /etc/os-release
    $ sudo yum -y install yum-utils
-   $ sudo -E VERSION_ID=$VERSION_ID yum-config-manager --add-repo "http://download.opensuse.org/repositories/home:/katacontainers:/release/CentOS_${VERSION_ID}/home:katacontainers:release.repo"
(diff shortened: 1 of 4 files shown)


grahamwhaley:
CI not happy - can you check if the OBS failures in the 16.04 logs are directly relevant/caused by this change @jodh-intel ?

adrianschroeter:
yes, it depends on the mirror (has nothing todo with OBS itself actually). Will beg our admins to review all of the configurations, but there might be mirrors where this couldn't get avoided due to lack of SSL.

AFAIK you should not have these problems when you start with http:// directly. relying on https is anyway not good security wise here, since nearly everybody can become a mirror.

So the real security is applied via the gpg signatures.

amshinde:
@jodh-intel Should we just stick to http urls?

jodh-intel:
Yep - seems we have to ;((

Closing...

jodh-intel:
Re-closing as (ateast) `apt` gets unhappy when it goes via a mirror which downgrades from https -> http.

PR #233, opened 2018-08-24
docs: Don't hardcode minimum golang version

### LINKED ISSUE #232: Don't hardcode golang version in devguide
Remove the hardcoded golang version in the developer guide since:

- We don't want to have to maintain that information in two places

  It's already in https://github.com/kata-containers/runtime/blob/master/versions.yaml.

- The hard-coded value is now stale

   See https://github.com/kata-containers/runtime/pull/643.

- The doc already provides the `versions.yaml` link which allows developers to determine the currently supported versions.

jodh-intel:
The minimum golang version should be defined *once* - in [1]. Since the
developer guide already provides a link to that human-readable file,
remove the hard-coded golang version number to avoid having to maintain
that part of the devguide.

Fixes #232.

[1] - https://github.com/kata-containers/runtime/blob/master/versions.yaml

### DIFF
--- a/Developer-Guide.md
+++ b/Developer-Guide.md
@@ -73,7 +73,7 @@ guest kernel.
 
 You need to install the following to build Kata Containers components:
 
-- [golang](https://golang.org/dl) version 1.8.3 or newer.
+- [golang](https://golang.org/dl)
 
   To view the versions of go known to work, see the `golang` entry in the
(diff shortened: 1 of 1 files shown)


nitkon:
LGTM

chavafg:
lgtm

chavafg:
Since this is a very tiny change and already has other 3 acks, will proceed to merge.

Issue #252, opened 2018-09-25
Limitations: document the current status of selinux support


grahamwhaley:
Selinux support is not presently available under Kata, as seen at https://github.com/kata-containers/runtime/issues/784.
At a minimum, document that in the limitations doc. It is probably wise to also open a specific issue, probably on the agent repo?, that selinux is not supported.
We need to start with that age-old discussion of where is one expecting it to be applied - on the host or inside the VM?

jodh-intel:
:chestnut: :chestnut: :chestnut: (:arrow_left: speaking of age-old... those are *chestnuts*! :smile:)

PR #253, opened 2018-09-25
Limitations: add selinux support limitation

grahamwhaley:
We do not currently support enablement of `selinux` in the
dockerd config. Document that.

Fixes: #252

### DIFF
--- a/Limitations.md
+++ b/Limitations.md
@@ -14,6 +14,8 @@
     * [Resource management](#resource-management)
         * [docker run and shared memory](#docker-run-and-shared-memory)
         * [docker run and sysctl](#docker-run-and-sysctl)
+    * [Docker daemon features](#docker-daemon-features)
+        * [selinux support](#selinux-support)
 * [Architectural limitations](#architectural-limitations)
     * [Networking limitations](#networking-limitations)
         * [Support for joining an existing VM network](#support-for-joining-an-existing-vm-network)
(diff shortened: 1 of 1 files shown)


nitkon:
LGTM

jodh-intel:
I appreciate you're documenting the docker server side, but is this all essentially the same as the following?:

- https://github.com/kata-containers/documentation/blob/master/Limitations.md#docker---security-opt-option-partially-supported

That section doesn't spell it all out (as we don't want to have to enumerate all the security types which may change), but we don't support `--security-opt label:$selinux_context`.

grahamwhaley:
Ah, so that is the link to how/where it gets passed into the runtime @jodh-intel - how about I add a ref in my new section over to that --security-opt one?

jodh-intel:
@grahamwhaley - yep, that plan sounds good to me.

amshinde:
@grahamwhaley k8s also has selinux labels: https://kubernetes.io/docs/tasks/configure-pod-container/security-context/#assign-selinux-labels-to-a-container
We should mention that as well for completeness.

grahamwhaley:
OK, I've added a link to the other section to ref the non-working security option, and also a link out to the k8s selinux label page. @jodh-intel @amshinde ptal.

jodh-intel:
lgtm

grahamwhaley:
afaik, this is done. In theory, needs one more ack from @kata-containers/documentation , but, it seems pullapprove may not have registered @jodh-intel ack. I'm OK with somebody pushing this through if they want.

GabyCT:
lgtm

PR #271, opened 2018-10-16
Update Ubuntu instructions to add snaps

sfabel:
Added instructions for using snaps to install the Kata Container runtime.

### DIFF
--- a/install/ubuntu-installation-guide.md
+++ b/install/ubuntu-installation-guide.md
@@ -1,6 +1,7 @@
-# Install Kata Containers on Ubuntu
+# Install Kata Containers on Ubuntu 
 
 1. Install the Kata Containers components with the following commands:
+ * Using debs
 
    ```bash
    $ ARCH=$(arch)
@@ -9,6 +10,10 @@
    $ sudo -E apt-get update
    $ sudo -E apt-get -y install kata-runtime kata-proxy kata-shim
    ```
+   
+ * Using snaps
+
+    Install Kata Containers using `snap install kata-containers --classic`.
 
 2. Decide which container manager to use and select the corresponding link that follows:
 


jodh-intel:
Hi @sfabel - Thanks for raising! I'm kind of amazed we forgot to create this before :smile:

Travis is failing and if you look at the log it shows:

```
Found 1 commit between commit 0d25a024341b4911634ba283ec8e59343052c667 and branch master
ERROR: Commit 0d25a024341b4911634ba283ec8e59343052c667: Failed to find subsystem in subject: "Update Ubuntu instructions to add snaps"
commits for the project.

```

This is caused by our [`checkcommits`](https://github.com/kata-containers/tests/tree/master/cmd/checkcommits) tool complaining there is no "subsystem" prefix in the first line of your commit. You also need to reference a github issue in one of your commits so I've raised #274 for you.

Don't panic though! :) The fix is simple - you just need to tweak your commit to something like the following, rebase and then re-push:

```
docs: Update Ubuntu instructions to add snaps

Added instructions for using snaps to install the Kata Container runtime.

Fixes #274.

Signed-off-by: Your Name <your-email@somewhere.com>
```

However, since snappy is not just confined to Ubuntu (it's available in the Fedora repo for example), I think it would actually make sense to create a new document called:

- `install/snappy-installation-guide.md`
  (or maybe that filename should contain `snap`, or `snapd` instead?)

Let's get some input from @devimc too as he created the snaps for Kata...

marcov:
Hi @sfabel, I am glad the documentation is getting better by the day :)

I also agree with the proposal of @jodh-intel of creating a dedicated page if possible, and link it the [main install page](https://github.com/kata-containers/documentation/blob/master/install/README.md).

A small issue I see in updating this page directly, is that any new code block added may be picked up by the [kata-doc-to-scritp.sh](https://raw.githubusercontent.com/kata-containers/tests/master/.ci/kata-doc-to-script.sh) to generate installation instructions for kata. So the generated installation script may contain a mix of different methods.

jodh-intel:
I think the CI should be fine since if any files below `install/` are changed, the following script is run:

- https://github.com/kata-containers/documentation/blob/master/.ci/test-install-docs.sh

That will look for the distro-specific install guide and test that it works. However, if a new snappy install guide is created, the script above will just re-check the existing distro install guides.

Any new docs will also be checked by https://github.com/kata-containers/tests/blob/master/.ci/static-checks.sh#L370 to ensure that the [code blocks](https://github.com/kata-containers/documentation/blob/master/Documentation-Requirements.md#code-blocks) look vaguely "sane". As that link show, the convention we follow is that ` ```bash ` blocks refer to code that the user should run but ` ```sh ` blocks are for generic shell code that users shouldn't necessarily run.

In this case, it doesn't really matter which is chosen as the CI won't try to execute the snappy install doc code blocks. However, we could change the CI to execute them once we tell the CI which distros support snappy.

marcov:
Thanks @jodh-intel, so as long as distro-specific install guides contain only a single way install kata it should all should be good.

jodh-intel:
@marcov - yes. In fact, we can support "conditional execution" in the distro install docs by having the alternative options use ` ```sh ` rather than ` ```bash `, but currently that means we only test the bash codepath. That should be sufficient and we should do everything we can to keep the install docs simple and "linear" (so the user doesn't need to make decisions when working through an individual install guide).

sfabel:
OK... So do you want me to do anything now? 😉 I'm getting confused...

jodh-intel:
Hi @sfabel - sorry about the confusion.

What do you think @devimc? Can we just move atleast parts of https://github.com/kata-containers/packaging/blob/master/snap/README.md to a file like `install/snappy-installation-guide.md` in this repo?

jodh-intel:
Re-ping @devimc.

install/ubuntu-installation-guide.md
```diff
@@ -1,6 +1,7 @@
-# Install Kata Containers on Ubuntu
+# Install Kata Containers on Ubuntu
```
devimc: please remove the extra space in the end of the line

install/ubuntu-installation-guide.md
```diff
@@ -1,6 +1,7 @@
-# Install Kata Containers on Ubuntu
```
devimc: Missing `Signed-off` in you commit message

devimc:
@jodh-intel sounds good

raravena80:
@sfabel ping, any updates? Thx!

sfabel:
@raravena80 @jodh-intel - not sure what went wrong? CI tells me the commit message is still in the wrong format but it looks good to me?

jodh-intel:
Hi @sfabel - it's complaining that your second commit doesn't have a [subsystem](https://github.com/kata-containers/community/blob/master/CONTRIBUTING.md#subsystem) (aka prefix). We require the first word of the first line of each commit to have prefix and a colon before the brief description. Here, "`docs:`" would make sense. However, I don't think you need two separate commits so I would `git rebase -i master` and then squash the commits together into a single one.

But since @devimc has agreed to moving https://github.com/kata-containers/packaging/blob/master/snap/README.md to this repo, you probably want to just copy that file to `install/snappy-installation-guide.md` and update the commit accordingly. Actually, when you do this, could you add a comment in the commit along the lines of:

> Note that this file was copied from https://github.com/kata-containers/packaging/blob/master/snap/README.md (commit https://github.com/kata-containers/packaging/commit/96ed56ec3fb4f3e095c1ad7ebe3c9feef6af63a7).

sfabel:
@jodh-intel - apologies for the silence on my part (travel). OK - seems like it might be easier to just close this PR, and I'll simply file a new one, creating the /snap/README.md from scratch?

jodh-intel:
Hi @sfabel - sure, something like copying `/snap/README.md` from the packaging repo to `install/snappy-installation-guide.md` in this repo.

PR #285, opened 2018-10-26
Use https instead of http in package installs

ttx:
Relying on bare HTTP to retrieve release keys or packages opens yourself to a variety of MITM attacks and root compromises.

I checked that download.opensuse.org is accessible using https://, so this should really be used instead.


jodh-intel:
Yep - we've been round this one before :smile:

See:

- https://github.com/kata-containers/documentation/pull/83#issuecomment-412438187
- https://github.com/kata-containers/documentation/issues/212

PR #286, opened 2018-10-26
Use https instead of http in package installs

ttx:
Relying on bare HTTP to retrieve release keys or packages
opens yourself to a variety of MITM attacks and root compromises.

I checked that download.opensuse.org is accessible using https://,
so this should really be used instead.

Signed-off-by: Thierry Carrez <thierry@openstack.org>

### DIFF
--- a/install/README.md
+++ b/install/README.md
@@ -54,7 +54,7 @@ Manual installation instructions are available for [these distributions](#suppor
 >  first read [the upgrading document](../Upgrading.md).
 
 > **Notes on releases**:
-> - [This download server](http://download.opensuse.org/repositories/home:/katacontainers:/releases:/)
+> - [This download server](https://download.opensuse.org/repositories/home:/katacontainers:/releases:/)
 > hosts the Kata Containers packages built by OBS for all the supported architectures.
 > Packages are available for the latest and stable releases (more info [here](https://github.com/kata-containers/documentation/blob/master/Stable-Branch-Strategy.md)).
 >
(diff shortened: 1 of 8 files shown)


jodh-intel:
See: https://github.com/kata-containers/documentation/pull/285#issuecomment-433400660.

ttx:
Missed that!

I somewhat disagree with those comments: while it is definitely possible to validate the keys externally, the instructions as posted (for example the 'curl http:// | sudo apt-key add -' posted for Ubuntu) do not include any check that the key matches what is expected and will silently accept any key, so the attack vector is still very much present...

But I agree that if some mirrors are not accessible using https:// my change is not the way to fix it.

ttx:
Follow-up on issue #212

jodh-intel:
@ttx - to be clear, we would *love* to be able to offer end-to-end https but OBS cannot guarantee that due to the mirrors being out of their control.

However, if we can ratchet up the security a bit, please do raise a PR to improve the docs we have!

Note that the "Automatic", "Manual" and "Scripted" installation methods shown in the table on this page are actually executed by the CI so we know they are correct:

- https://github.com/kata-containers/documentation/tree/master/install#installing-on-a-linux-system

Hence, if the docs could be updated in such a way that we could have the CI run any additional commands to validate the keys, etc, that would be ideal. If so, please use bash code blocks for code you want the CI to run:

- https://github.com/kata-containers/documentation/blob/master/Documentation-Requirements.md#code-blocks

/cc @marcov.

Issue #287, opened 2018-10-30
the cni diagram png is not found


liucimin:
https://github.com/kata-containers/documentation/blob/master/arch-images/CNI_diagram.png

the cni diagram png is not found?

PR #221, opened 2018-08-17
architecture.md: add network hotplug diagram

bergwolf:
To document the design we have agreed on network hotplug support.

Fixes: #220

cc @WeiZhang555 @caoruidong

PR #288, opened 2018-10-31
docs: Delete the wrong cni flow in README

liucimin:
docs: Delete the wrong cni flow in README

Fix the wrong  Network CNI section.

Fixes: #287 

Signed-off-by: Liu Cimin <lcmjkson@gmail.com>

### DIFF
--- a/architecture.md
+++ b/architecture.md
@@ -476,28 +476,6 @@ __Runtime network setup with CNM__
 5. Create bridge, TAP, and link all together with network interface previously
   created
 
-=======
-### CNI
-
-![CNI Diagram](arch-images/CNI_diagram.png)
-
-__Runtime network setup with CNI__
-
-1. Create the network namespace.
-
-2. Get CNI plugin information.
-
-3. Start the plugin (providing previously created network namespace) to add a network
-  described into `/etc/cni/net.d/ directory`. At that time, the CNI plugin will
-  create the `cni0` network interface and a veth pair between the host and the created
-  netns. It links `cni0` to the veth pair before to exit.
-
-4. Create network bridge, TAP, and link all together with network interface previously
-  created.
-
-5. Start VM inside the netns and start the container.
-
-
 ### Network Hotplug
 
 Kata Containers has developed a set of network sub-commands and APIs to add, list and


grahamwhaley:
Hi @liucimin - thanks for your contribution! :-)
Looking at the history of this bit of the document, I think we just need to check for some validity.
It looks like @amshinde removed this CNI section back on #235, and then @bergwolf added it back in at #221 (but not the png file). It's not clear to me if maybe we had a sequencing problem there with the PRs landing, so I'd like to ask @amshinde and @bergwolf to confirm if we still want this section, in which case we'll add the png file back in :-)

Other than that, the CI is complaining as your PR does not match our required format. Here is the output of the Travis CI logs, which then points you to the requirements as well :-). You may wish to hold off updating though until we confirm we do need that diagram back!

```
Found 1 commit between commit 92ac9fc4fa672150efdbbb6042750f170742fe46 and branch master
ERROR: Commit 92ac9fc4fa672150efdbbb6042750f170742fe46: Failed to find subsystem in subject: "'add_cni_diagram"
commits for the project.
```

bergwolf:
Good catch @liucimin and @grahamwhaley. I think it's my bad that I reintroduced the old texts when rebasing #221 after #235 is merged. @liucimin would you like to change the PR to remove the obsolete CNI section?

liucimin:
> Good catch @liucimin and @grahamwhaley. I think it's my bad that I reintroduced the old texts when rebasing #221 after #235 is merged. @liucimin would you like to change the PR to remove the obsolete CNI section?

Fine.
I think this pr again.And i will remove the obsolete CNI section.

But there will no cni flow in the readme.In my opion,the cni section should cri-o or contained to do?
And kata container dont need the CNI?
@bergwolf

bergwolf:
LGTM, thanks @liucimin !
 
> But there will no cni flow in the readme.In my opion,the cni section should cri-o or contained to do?
And kata container dont need the CNI?

kata containers do work with CNI. It's simply handled similarly as CNM. We scan the sandbox netns on the host, copy the network setup (interfaces and routes) from the netns to the guest. There is also a `kata-netmon` process that helps monitoring the sandbox netns changes after it is created.

I agree we need clearer documentation about this. @amshinde Are you planning on adding more docs about CNI handling after #235?

liucimin:
> kata containers do work with CNI. It's simply handled similarly as CNM. We scan the sandbox netns on the host, copy the network setup (interfaces and routes) from the netns to the guest. There is also a kata-netmon process that helps monitoring the sandbox netns changes after it is created.

Thanks for the reply.
But i feel confuse to this Explanation.I think the kata container is just runtime.And if the cri-o use the kata-container as runtime for untrust container,the cri-o will call cni to make the netns and other resouce such as veth-pair interface.So if then the cri-o to start the kata container,then the kata-container will call cni again?Is the duplicate? Or the kata-container will not call the same cni plugin , the kata-container call itself cni plugin?

Hope for your reply!Thx

bergwolf:
@liucimin Kata containers won't call any CNI plugins. From kata containers point of view, what we see is a netns that is already setup by cni. However the veth pair only connects the host network and the netns. We still need to connect the veth device to the guest's tap device. And this part is the same as CNM handling in kata containers.

There are a few cases w.r.t. how veth device is connected to the guest's tap device, according to the different [internetworking model configs](https://github.com/kata-containers/runtime/blob/master/cli/config/configuration.toml.in#L238) but the general idea is the same.

liucimin:
@bergwolf 
Thx for your reply.And this is also my understanding.

jodh-intel:
ltm

jodh-intel:
lgtm

PR #313, opened 2018-11-26
Azure install guide

egernst:


egernst:
@klynnrif @intelkevinputnam can you PTAL?  

This is a basic placeholder which will likely be augmented later.

install/azure-installation-guide.md
```diff
@@ -0,0 +1,18 @@
+# Install Kata Containers on Azure
+
+Kata Containers on Azure use nested virtualization to provide an identical installation
+experience to Kata on your preferred Linux distribution. 
+
+This guide assumes you have an Azure account set up and tools to remote into your virtual
```
jodh-intel: Nit: "... to remotely login to your..." would be clearer I think.

install/README.md
```diff
@@ -84,6 +84,7 @@ Manual installation instructions are available for [these distributions](#suppor
 
 ## Installing on a Cloud Service Platform
 * [Amazon Web Services (AWS)](aws-installation-guide.md)
+* [Azure](azure-installation-guide.md)
```
jodh-intel: Oh wait - shouldn't we say "Microsoft* Azure"?

egernst: I'm not sure -- this came from Sarah initially - I'll leave it to her discretion.  @scooley - WDYT?

install/azure-installation-guide.md
egernst: done

install/README.md
jodh-intel: All the other providers are listed, so I think for consistency alone we should specify the providing company name.

grahamwhaley: I vote we should add 'Microsoft' in the phrase - that is how it is called on their site, and they no doubt have some trademark or copyrights, so we should ack here.

grahamwhaley:
I've labelled this as WIP, as it obviously will get more updates, and the two commits can be folded as well before merge imo.

grahamwhaley:
have re-read - I'm happy - it is maybe deceptively 'too simple' :-)
merging...

Issue #300, opened 2018-11-16
Lost all of my old images if I follow the docker step


allenhaozi:
$ dir=/etc/systemd/system/docker.service.d
$ file="$dir/kata-containers.conf"
$ sudo mkdir -p "$dir"
$ sudo test -e "$file" || echo -e "[Service]\nType=simple\nExecStart=\nExecStart=/usr/bin/dockerd -D --default-runtime runc" | sudo tee "$file"
$ sudo grep -q "kata-runtime=" $file || sudo sed -i 's!^\(ExecStart=[^$].*$\)!\1 --add-runtime kata-runtime=/usr/local/bin/kata-runtime!g' "$file"
$ sudo systemctl daemon-reload
$ sudo systemctl restart docker

this will make my existed image lost, because I set custom graph dir

jodh-intel:
Hi @Allenhaozi - I think you are referring to https://github.com/kata-containers/documentation/blob/master/Developer-Guide.md#update-docker-configuration?

Unfortunately, running the commands in that guide are "at your own risk":

- https://github.com/kata-containers/documentation/blob/master/Developer-Guide.md#warning
- https://github.com/kata-containers/documentation/blob/master/Developer-Guide.md#assumptions

However, ideally we would either not modify the docker `ExecStart=`, or we would ensure the existing settings are retained. Were you running `dockerd --storage-driver=...` ?

/cc @grahamwhaley, @chavafg.

marcov:
One improvement is to detect any existing docker service file with:
```
sudo systemctl show -p FragmentPath docker.service
```
This is because unit files could be placed in `/etc/systemd`, `/usr/lib/systemd`, ..., and this should be taken into account to avoid masking out an existing unit file.

PR #319, opened 2018-12-14
dev-guide: detect & use existing docker unit file

marcov:
Only create a new docker unit file if no other existing unit files
are detected. Creating a new docker file when not necessary may mask out
existing docker daemon configurations.

Fixes: #300

### DIFF
--- a/Developer-Guide.md
+++ b/Developer-Guide.md
@@ -259,7 +259,7 @@ $ script -fec 'sudo -E USE_DOCKER=true ./image_builder.sh ${ROOTFS_DIR}'
 ```
 
 > **Notes:**
-> 
+>
 > - You must ensure that the *default Docker runtime* is `runc` to make use of
 >   the `USE_DOCKER` variable. If that is not the case, remove the variable
 >   from the previous command. See [Checking Docker default runtime](#checking-docker-default-runtime).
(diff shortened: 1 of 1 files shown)


marcov:
I agree, the page as it is now looks looks at best unappealing :) And still the code blocks need to be copy pasted line-by-line, because of the leading `$`

jodh-intel:
Yes, the dollars are there [for clarity](https://github.com/kata-containers/documentation/blob/master/Documentation-Requirements.md#code-blocks), but I know it's a pain to copy'n'paste as a result.

We could consider auditing the `install_*.sh` CI scripts to make them all safe to call outside of a CI environment:

- https://github.com/kata-containers/tests/tree/master/.ci

That would be delicate work as we'd need to ensure they don't destroy a developer system (so they would all need to potentially check for [`KATA_DEV_MODE`](https://github.com/kata-containers/tests#developer-mode). But we've also got:

- the [`kata-manager`](https://github.com/kata-containers/tests/tree/master/cmd/kata-manager) utility would could be expanded.
- *And* we've got https://github.com/kata-containers/tests/blob/master/cmd/container-manager/manage_ctr_mgr.sh.

So there is definitely an opportunity to do some refactoring imho ;)

/cc @chavafg, @GabyCT.

ydjainopensource:
@jodh-intel @marcov While we are changing this, why not add kata runtime into config.json. 

I have seen multiple issues arising because of users specifying gvisor in config.json and kata in systemd. Adding kata to config.json allows us to work well with the gvisor and other runtimes in general and ideally should be the way to go forward in my opinion.

ydjainopensource:
BTW, for parsing json in bash, there's a library called [jq](https://github.com/stedolan/jq) which we can use

marcov:
@ydjainopensource do you have a specific issue in mind? I can only see as a problem specifying  default runtimes in both or when picking the same name for a additional runtime.

If we move the config to the json file, there may be similar issues for users who already specified some options in the systemd unit file.

btw,

ydjainopensource:
@marcov The docker daemon does not allow a variable to be set in both through cli and daemon.json. This means if you specify kata in the cli and gvisor through daemon.json it would throw an error. So you would have to add kata in daemon.json. As for backward compatibility, yup this would be an issue. I can think of 2 ways to deal with this. 

1. we can maybe use the json alternative in the next major release.
1. we can have a post-install hook execute to migrate from systemd to json. We will have to be careful we don't change any other settings which the user might have changed there but otherwise this seems achievable.

btw, when is the next major release coming?

jodh-intel:
Ping @kata-containers/documentation ! This has been awaiting review for too long...

GabyCT:
lgtm

PR #327, opened 2018-12-27
vsock: Add introduction about VMWare guest

teawater:
Add introduction about use kata containers with vsock inside VMWare guest.

Fixes: #326

Signed-off-by: Hui Zhu <teawater@hyper.sh>

### DIFF
--- a/VSocks.md
+++ b/VSocks.md
@@ -94,6 +94,14 @@ $ sudo modprobe -i vhost_vsock
 The Kata Containers version must be greater than or equal to 1.2.0 and `use_vsock`
 must be set to `true` in the runtime [configuration file][1].
 
+### With VMWare guest
+To use Kata Containers with vsocks in a VMWare guest environment, first stop the vmware-tools service and unload the VMWare Linux kernel module.
+```
+sudo systemctl stop vmware-tools
+sudo modprobe -r vmw_vsock_vmci_transport
+sudo modprobe -i vhost_vsock
+```
+
 ## Advantages of using vsocks
 
 ### High density


VSocks.md
```diff
@@ -93,6 +93,13 @@ $ sudo modprobe -i vhost_vsock
 
 The Kata Containers version must be greater than or equal to 1.2.0 and `use_vsock`
 must be set to `true` in the runtime [configuration file][1].
+### With VMWARE host
+To use kata containers with vsock inside VMWARE host, need stop vmware-tools and unload VMWARE Linux kernel module.
```
gnawux: I think it is a VMWare guest, right?

gnawux:
much clearer now
lgtm

VSocks.md
```diff
@@ -93,6 +93,13 @@ $ sudo modprobe -i vhost_vsock
 
 The Kata Containers version must be greater than or equal to 1.2.0 and `use_vsock`
 must be set to `true` in the runtime [configuration file][1].
+### With VMWare guest
+To use kata containers with vsock inside VMWare guest, need stop vmware-tools and unload VMWare Linux kernel module.
+```
+sudo systemctl stop vmware-tools
+sudo modprobe -r vmw_vsock_vmci_transport
```
grahamwhaley: The change l g t m (ideally with the small suggested wording update from @jodh-intel applied).
I just want to check that we do have to unload the vmware vsock code before we can use the vhost-sock driver - I'm just wondering if there is anything that can be done at the code level in kata to choose which vsock transport we want etc. And the person I suspect would know is .... @stefanha

stefanha: Yes, vmw_vsock_vmci_transport needs to be unloaded on the host so that vhost_vsock can be loaded for Kata Containers.  This is a limitation of AF_VSOCK - it doesn't support nested virtualization.  Hopefully this limitation will be removed in the future but there is currently no active work in this area.

jodh-intel:
Thanks @teawater.

lgtm

Ping @kata-containers/documentation.

Issue #357, opened 2019-01-25
VERSION_ID in debian install can be unset under buster/sid(maybe testing too)


yyyeerbo:
VERSION_ID in buster/sid can be unset. default it to 9

Issue #344, opened 2019-01-11
Debian is no longer installable using kata-manager


jodh-intel:
From https://github.com/kata-containers/documentation/blob/master/install/installing-with-kata-manager.md#full-installation...

```
$ bash -c "$(curl -fsSL https://raw.githubusercontent.com/kata-containers/tests/master/cmd/kata-manager/kata-manager.sh) install-docker-system"
W: The repository 'http://ftp.debian.org unstable Release' does not have a Release file.
N: Data from such a repository can't be authenticated and is therefore potentially dangerous to use.
N: See apt-secure(8) manpage for repository creation and user configuration details.
E: Failed to fetch http://ftp.debian.org/dists/unstable/non-free/source/Sources  404  Not Found [IP: 130.89.148.12 80]
E: Some index files failed to download. They have been ignored, or old ones used instead.
gpg: no valid OpenPGP data found.
```

A bit of hacking resulted in the following (note: won't render using ` ```diff ` due to the backticks in the diff itself):

```

  diff --git a/install/debian-installation-guide.md b/install/debian-installation-guide.md
  index f031520..3c65818 100644
  --- a/install/debian-installation-guide.md
  +++ b/install/debian-installation-guide.md
  @@ -9,15 +9,14 @@
     
     ```bash
     $ sudo sh -c "echo '# for unstable packages
  -  deb http://ftp.debian.org unstable main contrib non-free
  -  deb-src http://ftp.debian.org unstable main contrib non-free' > /etc/apt/sources.list.d/unstable.list"
  +  deb http://ftp.debian.org/debian sid main' > /etc/apt/sources.list.d/unstable.list"
     ```
     
     Set the repository to a lower priority than stable, to ensures that APT will prefer stable packages over unstable ones. This can be specified in `/etc/apt/preferences.d/unstable`:
     
     ```bash
     $ sudo sh -c "echo 'Package: *
  -  Pin: release a=unstable
  +  Pin: release a=sid
     Pin-Priority: 10' >> /etc/apt/preferences.d/unstable"
     ```
   
  @@ -32,8 +31,8 @@
      ```bash
      $ ARCH=$(arch)
      $ source /etc/os-release
  -   $ sudo sh -c "echo 'deb http://download.opensuse.org/repositories/home:/katacontainers:/releases:/${ARCH}:/master/Debian_9/$(lsb_release -rs)/ /' > /etc/apt/sources.list.d/kata-containers.list"
[...]

PR #358, opened 2019-01-25
install: VERSION_ID in buster/sid can be unset. default it to 9

yyyeerbo:
Fixes: #357

Signed-off-by: Yang Bo <bo@hyper.sh>

### DIFF
--- a/install/debian-installation-guide.md
+++ b/install/debian-installation-guide.md
@@ -33,6 +33,9 @@
    $ export DEBIAN_FRONTEND=noninteractive
    $ ARCH=$(arch)
    $ source /etc/os-release
+   $ [ "$ID" = debian ] && [ -z "$VERSION_ID" ] && echo >&2 "ERROR: Debian unstable not supported.
+     You can try stable packages here:
+     http://download.opensuse.org/repositories/home:/katacontainers:/releases:/${ARCH}:/master" && exit 1
    $ sudo sh -c "echo 'deb http://download.opensuse.org/repositories/home:/katacontainers:/releases:/${ARCH}:/master/Debian_${VERSION_ID}/ /' > /etc/apt/sources.list.d/kata-containers.list"
    $ curl -sL  http://download.opensuse.org/repositories/home:/katacontainers:/releases:/${ARCH}:/master/Debian_${VERSION_ID}/Release.key | sudo apt-key add -
    $ sudo -E apt-get update


install/debian-installation-guide.md
```diff
@@ -32,7 +32,7 @@
    ```bash
    $ ARCH=$(arch)
    $ source /etc/os-release
-   $ sudo sh -c "echo 'deb http://download.opensuse.org/repositories/home:/katacontainers:/releases:/${ARCH}:/master/Debian_${VERSION_ID}/ /' > /etc/apt/sources.list.d/kata-containers.list"
+   $ sudo sh -c "echo 'deb http://download.opensuse.org/repositories/home:/katacontainers:/releases:/${ARCH}:/master/Debian_${VERSION_ID:=9}/ /' > /etc/apt/sources.list.d/kata-containers.list"
    $ curl -sL  http://download.opensuse.org/repositories/home:/katacontainers:/releases:/${ARCH}:/master/Debian_${VERSION_ID}/Release.key | sudo apt-key add -
```
amshinde: Add the default here as well ? @yyyeerbo

yyyeerbo: where? in curl -sL xxxx?
NO. VERSION_ID is set to 9 after sh -c executed, we don't have to set VERSION_ID again.

jodh-intel: This fixes the immediate problem, but is going to be fragile: we don't want to keep updating this doc as Debian moves it's stable release number. Looking at a Debian test box with Kata installed:

```sh
$ cat /etc/os-release 
PRETTY_NAME="Debian GNU/Linux buster/sid"
NAME="Debian GNU/Linux"
ID=debian
HOME_URL="https://www.debian.org/"
SUPPORT_URL="https://www.debian.org/support"
BUG_REPORT_URL="https://bugs.debian.org/"
```

That box *was* installed as Debian 9 (stretch). Looking at a vanilla stretch box:

```sh
$ cat /etc/os-release 
PRETTY_NAME="Debian GNU/Linux 9 (stretch)"
NAME="Debian GNU/Linux"
VERSION_ID="9"
VERSION="9 (stretch)"
ID=debian
HOME_URL="https://www.debian.org/"
SUPPORT_URL="https://www.debian.org/support"
BUG_REPORT_URL="https://bugs.debian.org/"
```

A quick look shows the first system was subsequently `apt-get dist-upgrade`'d rather than `apt-get upgrade`'d. But this is still a bit surprising to me - the system *was* stretch - where did we say we wanted it to become sid?

Hence, I think we have a problem with the code in this doc which adds the unstable (sid) repo:

- https://github.com/kata-containers/documentation/blob/master/install/debian-installation-guide.md

We simply want to install a single package from sid and *never* install any other package from that version.

yyyeerbo:
I see. The doc itself is for debian stable..

Do you mean there is unexpected upgrade from stable to unstable? I don't know that issue.  :)

jodh-intel:
> Do you mean there is unexpected upgrade from stable to unstable? 

Yes :) I certainly wasn't expecting to end up with a sid installation. Any thoughts @zeigerpuppy?

yyyeerbo:
For unstable/testing box, is it acceptable to use ${VERSION_ID:=unstable} and put a link(debian_unstable -> debian_&lt;latest version&gt;) in OBS?  Or use another repository in OBS for unstable?

jodh-intel:
@yyyeerbo - oh I see - you're actually running unstable normally?

Let's get input from @marcov on ^.

yyyeerbo:
> @yyyeerbo - oh I see - you're actually running unstable normally?

yes.

zeigerpuppy:
the docs I suggested were for Debian stable (Stretch = 9).  The unstable packages repo was only to `librdbd1`.

Hence the unstable repo is given a low priority to prevent pulling from there unless specifically requested buy the user.

in `/etc/apt/preferences.d/unstable`

```
Package: *
Pin: release a=unstable
Pin-Priority: 10
```

If the user is on Sid, then the version of `librdbd1` will be recent enough to just do a normal `apt install`

jodh-intel:
Thanks @zeigerpuppy. Is there a way to stop users getting upgraded to sid though? Or is it going to be a case of "just don't run `apt-get dist-upgrade`"?

zeigerpuppy:
both `apt-get upgrade` and `apt-get dist-upgrade` will be safe and stay on stable as long as the unstable packages have been set to a lower priority than the stable packages.

This can be simulated by running:

```bash
apt-get -s upgrade
apt-get -s dist-upgrade
```
these will produce similar results (not installing from unstable)

Contrast this with:

```bash
apt -s dist-upgrade -t unstable
```

which will install upgraded packages from unstable

jodh-intel:
Picking over my `/var/log/dpkg.log` and `/var/log/apt/history.log`, I still cannot see how my system got upgraded to sid but I can see I hit https://github.com/kata-containers/documentation/issues/344 and tried purge `docker-ce` and `librbd1` to reset. I'll assume for now the problem I'm seeing is user-error since nobody else has complained either ;)

However, regarding this PR, we have a few options if we detect an unset `VERSION_ID`. Before outlining the options, it's worth noting that although the [freedesktop `os-release` spec](https://www.freedesktop.org/software/systemd/man/os-release.html) says...

> The following OS identifications parameters **may** be set using os-release

Where the key word is "**may**", no other distro version simply omits the `VERSION_ID` afaik (does anyone know what Fedora's rawhide does?)

Further, since this is Debian unstable and we don't provide packages for that version... (because it's not stable and never going to be "released" :-) ... any changes we make to the install docs should be as minimal as possible imho.

So, the options:

- We could hard-code the Debian stable version

  Aka this PR. We have tried hard to avoid hard-coding *anything* *anywhere* in Kata so I don't think this is the way we should go.

- We *could* try to get the install doc to "resolve" the `VERSION_ID` itself

  The doc *could* do crazy things with `curl -l` to try to work out the latest version of Debian packages provided on OBS.
  - pros: It might work (today).
  - cons: It's a horrid hack, fragile and would only be used for a single distro version that we're not expecting folk to be running Kata on.

- We could just error out:
  ```sh
  $ [ "$ID" = debian ] && [ -z "$VERSION_ID" ] && echo >&2 "ERROR: Debian unstable not supported" && exit 1
  ```
  - pros: simple, and the user gets a helpful error message.
  - cons: Doesn't provide the best user experience. However, the OBS packages are built for Debian stable, not Debian unstable. If they work on sid, you got lucky, but all guarantees are out as we haven't built or tested for this version. Also, we don't expect many folk to ever hit this code path.

I think the error option is the right answer. If you are sophisticated enough to be running Debian unstable, you'll be able to work out what to do. Also, we could point the user at https://github.com/kata-containers/documentation/tree/master/install#scripted-installation which allows you to generate a script based on the install doc. You can then tweak that script as required to install on sid if you want to take that risk.

jodh-intel:
Thanks for the offer @marcov. However, my view is that although we want to be as "encompassing" as possible, we should be concentrating our packaging efforts on actual distro version *releases*. Debian sid is not a release, so doesn't qualify. But as I said above, if you're running sid, chances are you are a power user and will be able to use the scripted install method anyway. There is also apparently a security consideration with sid:

- https://www.debian.org/releases/sid/index.en.html

Of course, if we could find [someone](https://github.com/kata-containers/packaging/issues/6#issuecomment-453940053) to work on the existing [Debian packaging](https://github.com/kata-containers/packaging/tree/master/obs-packaging) and be an advocate for getting official packages accepted into Debian, sid would have access to Kata automatically.

In summary, I think we should:

- Continue to ramp up support for Debian stable in Kata.
- Not change OBS just for sid.
- Consider adding the error logic in the debian install doc to detect sid.

Just a few thoughts, but what do others think?

/cc @grahamwhaley, @egernst

yyyeerbo:
both error out and renaming to stable work for me, I am ok with them. we have to decide which way to go if no one has better idea.. :)

jodh-intel:
Thanks for the reminder @yyyeerbo.

Ping @kata-containers/documentation - Please could folk comment on the options here?

jodh-intel:
Ping @kata-containers/documentation - I'd like to get this PR resolved. I would still prefer we error out if the scripts detect a user attempting to install on sid as we don't test sid. See my comments above...

grahamwhaley:
I'm prone to vote 'abort with a nice error message for sid saying you need to power-install from source'.

yyyeerbo:
updated. Sorry for the delay..

Issue #371, opened 2019-02-04
zun guide: update kata install URL


marcov:
The old "release/" subproject on OBS [is going to be nuked](https://github.com/kata-containers/packaging/issues/325) soon, update the guide to just point to the Kata manual installation instructions for Ubuntu.

PR #372, opened 2019-02-04
zun: update kata install instructions

marcov:
Update zun guide to point to the kata installation instructions for
Ubuntu.

Fixes: #371

Signed-off-by: Marco Vedovati <mvedovati@suse.com>

### DIFF
--- a/zun/zun_kata.md
+++ b/zun/zun_kata.md
@@ -50,14 +50,10 @@ zun logs test
 zun delete test
 ```
 
-## Install `kata-runtime`, `kata-shim`, and `kata-proxy`
+## Install Kata Containers
 
-```
-sudo sh -c "echo 'deb http://download.opensuse.org/repositories/home:/katacontainers:/release/xUbuntu_$(lsb_release -rs)/ /' >> /etc/apt/sources.list.d/kata-containers.list"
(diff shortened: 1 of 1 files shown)


marcov:
Mmmh, now I have the doubt that the installation instructions **needs** to be hardcoded to be machine parsed. In this case I can update the PR.

/cc @jodh-intel @grahamwhaley

grahamwhaley:
@marcov - I'm not sure I follow. the PR here looks good, which drops the hand coding of `$(lsb_release -rs)` for instance, and moves the onus over to the ubuntu install dirs. Or, are you saying you could just say 'install per the OS of your choice', and not even explicitly mention Ubuntu? Which, would also work for me.

marcov:
@grahamwhaley I was wondering if this page could parsed by something similar to kata-doc-to-script to generate some install script. In that case a link wouldn't work.

jodh-intel:
@marcov - we could do that, but we'd probably need to put the logic into https://github.com/kata-containers/tests/blob/master/cmd/kata-manager/kata-manager.sh itself (as we do for the docker install atm).

lgtm

jcvenegas:
ping, this need rebase master.

grahamwhaley:
:-) I think I'm going to go disable all those 'branch is out of date' warnings from github, as most of the time I think they are still mergeable - and github gives us a different (and more blatant) warning if they are not mergeable..

marcov:
Rebased

/test

GabyCT:
lgtm

PR #397, opened 2019-03-08
docs: Add RHEL and SLES install warnings

### LINKED ISSUE #396: Add RHEL install guide warning that packages not tested
See: https://github.com/kata-containers/packaging/pull/370#issuecomment-470189028

jodh-intel:
Unfortunately, at present we have no way of testing Kata packages for Red Hat Enterprise Linux (RHEL) or SUSE Linux Enterprise (SLES).

Add warnings to the RHEL and SLES install guides explaining this and advising users to exercise caution. Hopefully, we will be able to drop this warning soon (either when we have the ability to test on RHEL/SLES or when Kata packages are available in RHEL/SLES).

Fixes #396.

### DIFF
--- a/install/rhel-installation-guide.md
+++ b/install/rhel-installation-guide.md
@@ -1,5 +1,13 @@
 # Install Kata Containers on RHEL
 
+> **Warning:**
+>
+> - The RHEL packages are provided as a convenience to users until native
+>   packages are available in RHEL. However, they are **NOT** currently tested
+>   (although CentOS is) so caution should be exercised.
+>
+>   See https://github.com/kata-containers/ci/issues/3 for further details.
+
 1. Install the Kata Containers components with the following commands:
 
    ```bash
--- a/install/sles-installation-guide.md
+++ b/install/sles-installation-guide.md
@@ -1,5 +1,13 @@
 # Install Kata Containers on SLES
 
+> **Warning:**
+>
+> - The SLES packages are provided as a convenience to users until native
+>   packages are available in SLES. However, they are **NOT** currently tested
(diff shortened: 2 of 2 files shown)


jodh-intel:
As I said in the commit message, let's hope this warning will only be required temporarily.

Ping @kata-containers/documentation.

/cc @ttx, @lsm5, @fatherlinux, @mrguitar.

fatherlinux:
@jodh-intel I don't fully understand what is going on here, but having some people take a look. Sadly, I don't know their user names in GitHub, or I would just tag them in. Are we sure the gcc feature you are looking for isn't back ported in RHEL? Why is it working in CentOS, but not RHEL? That part is confusing to me?

jodh-intel:
Hi @fatherlinux - the problem we have as I understand it is that the Kata project doesn't have a way to obtain RHEL licenses. So although we can build packages on RHEL thanks to OBS, we cannot test them, hence the need for this PR.

I think we may have tried a developer / trial version of RHEL way back, but that didn't provide package updates / full RHEL repo access I think. There may also have been an issue trying to use RHEL in our CI environment (Jenkins) - does that sound right to you @chavafg, @gabibeyer, @grahamwhaley?

There is also a bit of detail on https://github.com/kata-containers/ci/issues/3.

I've just realised that we have the same issue with SLES, so I've updated this PR for that too.

jodh-intel:
Ping @kata-containers/documentation.

jodh-intel:
Ping @kata-containers/documentation - small change set seeks sponsors for quick review :smile:

PR #402, opened 2019-03-18
Add links to all how-to and use-case docs in README.md

amshinde:
Add links to all how-to and use-case docs in README.md so that they can be navigated to. Currently there is no way to reach them except actually browsing through the repo.


gnawux:
The page itself is a TOC actually 😂  

Let's merge it firstly

Issue #417, opened 2019-03-29
improve organization of documentation repo


egernst:
We could use some shuffling/cleanup.
- [x] move Zun and its artifacts to use cases: https://github.com/egernst/documentation/commit/39cd6aa62603e932243290efc026ed3463371e07
- [ ] reference constraints document from architecture/design mds
- [x] move constraints.md under design folder, reference from architecture.md?
- [x] move Vsocks.md to design folder, reference from architecture.md?
- [ ] create a processes folder, and move relevant .mds under here
- [ ] review top level of README for usefullness. 
- [ ] need to make sure each .md is referenced at least once

egernst:
/cc @eadamsintel @devimc 
@gabibeyer @ganeshmaharaj  - as new(er) contributors to the code base, I could use your help in viewing our design/architecture docs and providing some fresh insights into usability.

/cc @jodh-intel  @grahamwhaley FYI.

egernst:
@mimeriza  - if you have cycles, this may actually work well for you, too.

eadamsintel:
I would consider combiningg the "how to" folder with the "use case" folder.  You could call it "user guides"  It is not obvious which folder a user would want to click on.  I might think how do I use this with a GPU device and you would have to look in the "use case" folder.  There should probably be a rather large guide on how to use kata with kubernetes that goes into runtimeclass, how to use kata-deploy, etc.. and it should appropriately reference the architecture guide when talking about networking or storage and how it works with kubernetes.  The kubernetes information is spread throughout the other guides but users that just want to jump in and get started wouldn't care necessarily about other design aspects of kata.  As part of this I would move kata-deploy instructions to the install folder for people that would think to go there for kubernetes integration instructions. This can be referenced by a kubernetes guide in the how to section.

jodh-intel:
Related: https://github.com/kata-containers/tests/issues/1316.

PR #420, opened 2019-04-01
doc: update architecture.md link

gabibeyer:
update architecture.md link, since it has moved to within the
design/ directory.

contributes to: #417

Signed-off-by: Gabi Beyer <Gabrielle.n.beyer@intel.com>


egernst:
I wish we had a contributes-to.  Can we configure something like this in addition to "fixes" @jodh-intel ?

jodh-intel:
@egernst - How would we want this to work? "`Contributes-To:`" implies "partially fixes". We could update https://github.com/kata-containers/tests/tree/master/cmd/checkcommits to accept a "`Contributes-To: ###`" rather than a "`Fixes: ###`". I guess that would be sufficient since as GitHub won't understand that tag, it won't close the issue when the PR lands.

Issue #423, opened 2019-04-08
installation: Allow define branch


jcvenegas:
Installation steps should include variable to
define what kata branch want to use.

The installation steps only point to master.

PR #424, opened 2019-04-08
install: Allow define kata branch.

jcvenegas:
Kata OBS repositories provide multiple branch support.

Let define a variable to allow users define kata branch to use.

Fixes: #423


jcvenegas:
ping @kata-containers/documentation

PR #428, opened 2019-04-12
resource management: document current status, suggested fixes

egernst:
Let's use this document to drive the conversation around design changes to better manage Kata Containers' resources.

### DIFF
--- a/design/resource-management.md
+++ b/design/resource-management.md
@@ -1,0 +1,663 @@
+# cgroup updates in Kata
+
+## Background
+
+With 1.6 release of Kata Containers there are some issues with resource management resulting in inconsistent
+behavior.  This document descibes the state of 1.6, and a suggested implementation for 1.7 version of Kata.
+
+## Existing Behavior (Kata 1.6):
+
+### In Docker
+### In CRI-O
+### In Containerd
+
+
+## Proposed Changes
+
+### Alternatives Considered
+
+
+## Opens 
+
+
+Cgroup Updates in Kata
+
+
+
+# Backup / to be handled:
+
+
+## Containerd Handling Today
+
+The hierarchy and cgroup handling seems pragmatic in the case of containerd. The
+container cgroups are currently placed under the podcgroup.
+
+Output from containerd guaranteed pod, with two containers:
+
+```
+kata@kata-k8s-containerd:/sys/fs/cgroup/cpu,cpuacct/kubepods$ tree pode3bb46f0-5c99-11e9-903b-000d3a6d0876/
+pode3bb46f0-5c99-11e9-903b-000d3a6d0876/
+├── 47c6122142c76db9f2929c23e8693c641559cb6588fbf5e2f5c21623f5af2fd1
+│   ├── cgroup.clone_children
+│   ├── cgroup.procs
+│   ├── cpu.cfs_period_us
+│   ├── cpu.cfs_quota_us
+│   ├── cpu.shares
+│   ├── cpu.stat
+│   ├── cpuacct.stat
+│   ├── cpuacct.usage
+│   ├── cpuacct.usage_all
+│   ├── cpuacct.usage_percpu
+│   ├── cpuacct.usage_percpu_sys
+│   ├── cpuacct.usage_percpu_user
+│   ├── cpuacct.usage_sys
+│   ├── cpuacct.usage_user
+│   ├── notify_on_release
+│   └── tasks
+├── a18373fc91cf91d9b362a536afaecd0712a858bb992777d956948911a6d3b248
+│   ├── cgroup.clone_children
+│   ├── cgroup.procs
+│   ├── cpu.cfs_period_us
+│   ├── cpu.cfs_quota_us
+│   ├── cpu.shares
+│   ├── cpu.stat
+│   ├── cpuacct.stat
+│   ├── cpuacct.usage
+│   ├── cpuacct.usage_all
+│   ├── cpuacct.usage_percpu
+│   ├── cpuacct.usage_percpu_sys
+│   ├── cpuacct.usage_percpu_user
+│   ├── cpuacct.usage_sys
+│   ├── cpuacct.usage_user
+│   ├── notify_on_release
+│   └── tasks
+├── cgroup.clone_children
+├── cgroup.procs
+├── cpu.cfs_period_us
+├── cpu.cfs_quota_us
+├── cpu.shares
+├── cpu.stat
+├── cpuacct.stat
+├── cpuacct.usage
+├── cpuacct.usage_all
+├── cpuacct.usage_percpu
+├── cpuacct.usage_percpu_sys
+├── cpuacct.usage_percpu_user
+├── cpuacct.usage_sys
+├── cpuacct.usage_user
+├── f90eb972d44634f85726552a818578c4f28085f9a2f755fb8b18c402fe9cef6d
+│   ├── cgroup.clone_children
+│   ├── cgroup.procs
+│   ├── cpu.cfs_period_us
+│   ├── cpu.cfs_quota_us
+│   ├── cpu.shares
+│   ├── cpu.stat
+│   ├── cpuacct.stat
+│   ├── cpuacct.usage
+│   ├── cpuacct.usage_all
+│   ├── cpuacct.usage_percpu
+│   ├── cpuacct.usage_percpu_sys
+│   ├── cpuacct.usage_percpu_user
+│   ├── cpuacct.usage_sys
+│   ├── cpuacct.usage_user
+│   ├── notify_on_release
+│   └── tasks
+├── notify_on_release
+└── tasks
+
+```
+
+I suggest that we *do not* create cgroups with names identical to the containers, and instead create a single cgroup, sized appropriately, to constrain our hypervisor.
+
+
+## CRI-O
+
+`insert example here?
+`
+## Docker
+
+`$ docker run --cpus=3 -it busybox  sh `
+
+containerID: `9621fa5988bdd7ba5128f9530a618aa270f9425137e6ffb4d207e5329be9b3f4`
+
+```
+eernst@eernstworkstation:/sys/fs/cgroup/cpu,cpuacct/docker$ tree
+.
+├── 9621fa5988bdd7ba5128f9530a618aa270f9425137e6ffb4d207e5329be9b3f4
+│   ├── cgroup.clone_children
+│   ├── cgroup.procs
+│   ├── cpuacct.stat
+│   ├── cpuacct.usage
+│   ├── cpuacct.usage_all
+│   ├── cpuacct.usage_percpu
+│   ├── cpuacct.usage_percpu_sys
+│   ├── cpuacct.usage_percpu_user
+│   ├── cpuacct.usage_sys
+│   ├── cpuacct.usage_user
+│   ├── cpu.cfs_period_us
+│   ├── cpu.cfs_quota_us
+│   ├── cpu.shares
+│   ├── cpu.stat
+│   ├── notify_on_release
+│   └── tasks
+├── cgroup.clone_children
+├── cgroup.procs
+├── cpuacct.stat
+├── cpuacct.usage
+├── cpuacct.usage_all
+├── cpuacct.usage_percpu
+├── cpuacct.usage_percpu_sys
+├── cpuacct.usage_percpu_user
+├── cpuacct.usage_sys
+├── cpuacct.usage_user
+├── cpu.cfs_period_us
+├── cpu.cfs_quota_us
+├── cpu.shares
+├── cpu.stat
+├── notify_on_release
+└── tasks
+
+```
+
+
+
+## Kata implementation
+1. Move qemu and proxy from conmon cgroup to a new cgroup (kata-sandbox)
+	1.1 - The kata-sandbox cgroup must be a child of the parent cgroup, that is specified in the OCI spec
+	1.2 - The constraint for the kata-sandbox cgroup should be equal to -1 (no constraints), that way it inherits the constraint of its parent
+2. Only next cgroups will be honored:
+	* cpu
+	* cpuset: cpuset  initially  is a join of all cpusets?  
+3. Qemu vcpu threads shouldn't be moved into the sandbox cgroup, since the whole Qemu pid is moved into the kata-sandbox.
+4. Kata-shim ???
+
+## Appendix
+
+### Guaranteed YAML:
+```
+apiVersion: v1
+kind: Pod
+metadata:
+  name: guar-runc
+spec:
+  containers:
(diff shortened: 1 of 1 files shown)


design/resource-management.md
```diff
@@ -0,0 +1,329 @@
...
+
+##### Where are the v2-shim, QEMU, and Vhost processes
+
+As shown below, all of these are placed under the containerd.service system.slice.  For `containerd-shim-kata-v2`
+this isn't a major concern, as it is not expected to take much resource, and it is pretty closely
+coupled to containerd.
+
+QEMU itself and its vhost threads are very problematic.  Depending on the workload, these can consume a
+non-negligible amount of resources. Note, in the Kata implementation, these components are purposefuly
+not added to the constrained cgroup, the pause container. As a result, they fall under the caller's
```
bergwolf: you mean "not added to the pod cgroup"?

design/resource-management.md
```diff
@@ -0,0 +1,329 @@
...
+
+For containerd, the memory is being charged to containerd, which is basically unbounded.  This is bad for
+node stability, as the `pod` is essentially unbounded.
+
+## Proposed Changes
+
+### Summary
+ * Pause cgroup cpu shaes should be setup correctly.
+ * Do not create container cgroups on the host. Instead, create a pod sandbox group that is entirely managed by Kata
+ * Move the QEMU threads into the sandbox cgroup
```
bergwolf: Plus shimv2, kata-shim/proxy, and vhost threads? I think they all belong to the sandbox cgroup.

design/resource-management.md
```diff
@@ -0,0 +1,329 @@
...
+
+With these changes, performance and constraints for a pod is consistent. This constraining change will be
+more restrictive relative to existing design.
+
+The overheads associated with running a sandbox should be accounted for explicitly, and at the pod level.
+Once the Pod Overhead KEP is available, this should become a part of RuntimeClass, applied to pods which
+utilize the applicable RuntimeClass.  See [Pod Overhead KEP](https://github.com/kubernetes/enhancements/blob/master/keps/sig-node/20190226-pod-overhead.md)
+for more details.
```
bergwolf: Yes! We really need pod overhead to let kubernetes be aware of kata.

design/resource-management.md
```diff
@@ -0,0 +1,329 @@
...
+ * The pod sandbox cgroup should always be the summation of all container group resources.
+ * In the case of cri-o where it creates other cgroups for conmon, they will be siblings.
+ * The conmon cgroups today are rather large, so if conmon goes wild there is a possibility
+ the workload will get fewer resources. But it will not introduce any other side effects.
+
+## Alternatives Considered
+
+### Only constrain vCPUs, leaving remaining threads for system reserved
+
+This will, in some cases, provide improved performance.
```
bergwolf: The problem with system reserved is that it does not scale. So I don't think we should put scaling factors (qemu main thread, vhost etc.) there and it can result in unexpected system failures due to enforced constraints in system reserved.

bergwolf:
@egernst The other part of the cgroups handling is how we handle them in the guest. Do you think we should include it in this document?

design/resource-management.md
```diff
@@ -0,0 +1,329 @@
...
+and CRI-O.
+
+#### In Docker
+
+#### In Kubernetes + Containerd
+
+For each container in the pod, a cgroup is created within the pod cgroup (ie, under `/sys/fs/cgroup/*/kubepod/pod.*/`
+for a guaranteed pod). This is not necessary; only a single cgroup which constrains the hypervisor
+appropriately is required.
+
```
mcastelino: There is a cgroup for the pod itself. Within the cgroup of the pod child cgroups are setup. There is one for the pause container and one each for each container in the pod. Only the actual container processes are placed in these cgroups. The container framework processes like shim are placed under containerd systemd slice cgroup and as such are accounted for. This can be a potential issue is the framework ends up consuming resources. example logs, stdio.

design/resource-management.md
```diff
@@ -0,0 +1,329 @@
...
+##### Where are all the vCPUS
+
+The vCPUs are placed under the pause container:
```
mcastelino: Kata tries to setup the pause cgroup to represent the VM itself. It scales its resources to match the total resources assigned to the pod.

design/resource-management.md
```diff
@@ -0,0 +1,329 @@
...
+```bash
+# cat */cpu.cfs_quota_us
+300000
+200000
+500000
+```
+One drawback of this is that it assumes the existence of a pause container. This is an assumption
+based on the implementation of containerd/cri-o.  On the plus side, the cgroup is placed directly under
+the pod cgroup, which is created and managed by Kubelet.  Overall, this isn't terrible.
+
```
mcastelino: However CPU resources are proportional.  Hence if any of the sibling cgroups consume resources the summation logic does not work.

design/resource-management.md
```diff
@@ -0,0 +1,329 @@
...
+root@kata-k8s-containerd:/sys/fs/cgroup/cpu,cpuacct# grep -ir 24010
+system.slice/containerd.service/cgroup.procs:24010
+system.slice/containerd.service/tasks:24010
+```
+
+#### In Kubernetes  + CRI-O (v1 shim)
+
+CRI-O is very similar the containerd, except for where the non-constrained processes end up. Instead
+of being called by CRIO directly, kata-runtime is called from a process `conmon`, which is located
+in a cgroup under the pod-cgroup. As expected based on prior exapmles, cgroups are ceated for each
```
mcastelino: typos here.

design/resource-management.md
```diff
@@ -0,0 +1,329 @@
...
+{qemu-system-x86}(18281)
+{qemu-system-x86}(18368)
+{qemu-system-x86}(18369)
+{qemu-system-x86}(18370)
+```
+
+The QEMU, vhost, proxy and shim threads, however, are placed under the caller's cgroup, which in this case
+is `conmon`, which is a peer of the container cgroups we created. So, the good news is that QEMU, vhost,
+etc, are constrained within the pod's cgroup. The bad news is these will be constrained based on the values
+associated with conmon:
```
mcastelino: This is also wrong as conmon should be constrained with the absolute minmum resources it can get away with. Typically this is set to 2 (i.e. so that it does not impact the container scheduling).

design/resource-management.md
```diff
@@ -0,0 +1,329 @@
...
+This will, in some cases, provide improved performance.
+
+## Opens
+
+### static CPU configurations
+
+If static CPU policies are introduced, the end user will assign CPUs to a specific container within the pod.  Running
+IO threads on this CPU may not necessarily be desirable, compared to the users expectations.
+
+Long term (ie, with RuntimeClass augmented to handle pod overheads), we should create a seperate `cpuset` cgroup,
```
mcastelino: The cpuset cgroup is separate from CPU cgroup. Also upstream kubernetes is planning to eliminate the cfs quotas for containers with cpusets. So we really cannot place QEMU outside the cpuset.
https://github.com/kubernetes/kubernetes/issues/70585

design/resource-management.md
```diff
@@ -0,0 +1,329 @@
...
+due to the CPU constraints applied to conmon (where the io threads run).
+
+For containerd, the memory is being charged to containerd, which is basically unbounded.  This is bad for
+node stability, as the `pod` is essentially unbounded.
+
+## Proposed Changes
+
+### Summary
+ * Pause cgroup cpu shaes should be setup correctly.
+ * Do not create container cgroups on the host. Instead, create a pod sandbox group that is entirely managed by Kata
```
jcvenegas: Would be good to describe why: This is related with the cadvisor issue right ?

jcvenegas:
PR to move kata assets to its sandbox level cgroup based in sandbox cgroup  container. See: https://github.com/kata-containers/documentation/pull/428

design/resource-management.md
```diff
@@ -0,0 +1,329 @@
...
+
+## Opens
+
+### static CPU configurations
+
+If static CPU policies are introduced, the end user will assign CPUs to a specific container within the pod.  Running
+IO threads on this CPU may not necessarily be desirable, compared to the users expectations.
+
+Long term (ie, with RuntimeClass augmented to handle pod overheads), we should create a seperate `cpuset` cgroup,
+`kata-sandbox-vcpus`, alongside the standard sandbox cgroup, `kata-sandbox`. These would be siblings underneath the
```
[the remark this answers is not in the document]
egernst: @jcvenegas, these would need to be under parent (appropraite hierarchical location; where the caller expects it to be).  In case of kubernetes, under the pod-*.

I'm not sure we need sandbox-id in the naming; determining which sandbox it is associate with is determined by its hierarchical location.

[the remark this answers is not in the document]
mcastelino: @jcvenegas do we need the sandbox id? Can we not call it just sandbox in the case of crio and containerd with kubernetes.

@egernst In case of docker who creates the cgroup `/docker/3cca57349adcfb17f917d2312b568faa85515e6d5a8327a30ea8e5ecd47b0e1f`

Can we be under that.

design/resource-management.md
mcastelino: No. This is because it is not for the pause process. It is actually for the VM. Hence we cannot just add up and apply to pause. If we create the container cgroup and something runs in them as cgroups in the case of cpu is proportional we will get wrong behavior. So it is equally important not to create the container cgroups.
The conmon cgroup will be really tiny and hence should not effect the sandbox. (like how pause is setup by runc)

design/resource-management.md
```diff
@@ -0,0 +1,393 @@
+# cgroup updates in Kata
+
+  * [Background](#background)
+  * [Existing Behavior (Kata 1.6):](#existing-behavior--kata-16--)
+    + [Behavior Observed using various upper layer tools](#behavior-observed-using-various-upper-layer-tools)
+      - [In Docker](#in-docker)
+      - [In Kubernetes + Containerd](#in-kubernetes---containerd)
+        * [Where are all the vCPUS](#where-are-all-the-vcpus)
+        * [Where are the v2-shim, QEMU, and Vhost processes](#where-are-the-v2-shim--qemu--and-vhost-processes)
```
chavafg: broken link?

design/resource-management.md
```diff
@@ -0,0 +1,393 @@
...
+      - [In Kubernetes  + CRI-O (v1 shim)](#in-kubernetes----cri-o--v1-shim-)
```
chavafg: broken link?

design/resource-management.md
```diff
@@ -0,0 +1,393 @@
...
+    + [Issues with current implementation](#issues-with-current-implementation)
+      - [Accurate usage accounting](#accurate-usage-accounting)
+      - [Node stability](#node-stability)
+      - [Consistent guaranteed pod behavior](#consistent-guaranteed-pod-behavior)
+      - [OOM, unbound CPU utilization](#oom--unbound-cpu-utilization)
```
chavafg: broken link?

design/resource-management.md
```diff
@@ -0,0 +1,393 @@
...
+  * [Proposed Changes](#proposed-changes)
+    + [Summary](#summary)
+    + [Details](#details)
+      - [Pod Sandbox Cgroup](#pod-sandbox-cgroup)
+  * [Alternatives Considered](#alternatives-considered)
+    + [Only constrain vCPUs, leaving remaining threads for system reserved](#only-constrain-vcpus--leaving-remaining-threads-for-system-reserved)
```
chavafg: broken link?

design/resource-management.md
```diff
@@ -0,0 +1,393 @@
...
+  * [Opens](#opens)
+    + [static CPU configurations](#static-cpu-configurations)
```
chavafg: start with Capital letter? `[Static....]`

design/resource-management.md
```diff
@@ -0,0 +1,393 @@
...
+
+## Background
+
+With 1.6 release of Kata Containers there are some issues with resource management resulting
+in inconsistent behavior. This document descibes the state of 1.6, and a suggested implementation
+for 1.7 version of Kata.
+
+Before diving into the gaps and behavior exhibited in Kata Containers, it is important to have
+a thorough understanding of how cgroups are leveraged by Kubernetes.  It is both straight forward
+and confusing.  An in-depth guide is available for background in [mcastelino's gist](https://gist.github.com/mcastelino/b8ce9a70b00ee56036dadd70ded53e9f).
```
chavafg: remove extra-space?

jodh-intel:
@egernst - you could try running https://github.com/kata-containers/tests/pull/1542 over this PR as a useful test as it should detect the issues (and suggest fixes where possible).

raravena80:
@egernst any updates? Thx!

design/resource-management.md
grahamwhaley: I'd love us to not have to link to a gist of @mcastelino - they just feel too ephemeral to me - any chance we can get that info into the kata repos/docs??

grahamwhaley:
I skimmed over this - this looks like a good very technical doc. for us to have in Kata @egernst - do you plan to clean/land it?

caoruidong:
@egernst

jodh-intel:
Commit needs a tweak:

```
ERROR: Commit 82e4800b2796bd3760851f0892c10ad60618ac7b: Failed to find subsystem in subject: "resource management: document current status, suggested fixes"
commits for the project.
```

I suggest changing it to

> `resource-management: document current status, suggested fixes`.

design/resource-management.md
```diff
@@ -0,0 +1,393 @@
+# cgroup updates in Kata
+
+  * [Background](#background)
+  * [Existing Behavior (Kata 1.6):](#existing-behavior--kata-16--)
```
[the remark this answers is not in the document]
jodh-intel: Once the commit message is updated and re-pushed, the CI should detect all these broken links (and tell you what they should be ;)

design/resource-management.md
```diff
@@ -0,0 +1,393 @@
...
+      - [Pod Sandbox Cgroup](#pod-sandbox-cgroup)
+  * [Alternatives Considered](#alternatives-considered)
+    + [Only constrain vCPUs, leaving remaining threads for system reserved](#only-constrain-vcpus--leaving-remaining-threads-for-system-reserved)
+  * [Opens](#opens)
+    + [static CPU configurations](#static-cpu-configurations)
+
+## Background
+
+With 1.6 release of Kata Containers there are some issues with resource management resulting
+in inconsistent behavior. This document descibes the state of 1.6, and a suggested implementation
```
jodh-intel: We'll be releasing 1.8 soon so will this document still be relevant (since 1.6 won't be updated any further)?

jodh-intel:
@egernst - Do you think you'll get cycles to update this doc this week? Would be great to see this land.

jodh-intel:
Re-ping @egernst.

PR #446, opened 2019-04-25
improve developer guide doc

knightXun:
`Warning` is not good, User may become developer one day. `warning` is too scaring

### DIFF
--- a/Developer-Guide.md
+++ b/Developer-Guide.md
@@ -1,4 +1,4 @@
-* [Warning](#warning)
+* [Hint](#Hint)
 * [Assumptions](#assumptions)
 * [Initial setup](#initial-setup)
 * [Requirements to build individual components](#requirements-to-build-individual-components)
@@ -54,7 +54,7 @@
         * [Create an OCI bundle](#create-an-oci-bundle)
         * [Launch the runtime to create a container](#launch-the-runtime-to-create-a-container)
 
-# Warning
+# Hint
 
 This document is written **specifically for developers**: it is not intended for end users.
 


grahamwhaley:
Hi @knightXun - thanks for your contribution!

Personally, I think maybe we could say **Note**, which to me sounds better than **Hint**, and still less scary 😨 than **Warning** :-)

You will need to change your commit a little to make it conform with our CI format requirements as well. You can find details at this link: https://github.com/kata-containers/community/blob/master/CONTRIBUTING.md#patch-format
There is also a 'first example PR' section that may also help: https://github.com/kata-containers/community/blob/master/CONTRIBUTING.md#first-pr-example

Off the top of my head, the minimum things you will need to do:
- open an Issue so you can add a 'Fixes' to your commit
- Add a subsection item to the front of your subject lines (eg. `docs: improve.....`)
- Sign off the commit (`git commit -s` etc.).

And you can use `git commit --amend` to edit your commit and then `git push -f ....` to re-push your modified commit to this PR - you *do not* have to open a new PR, you can just update this one (sorry if that is obvious, we see quite a lot of first time committers closing and duplicating PRs when they don't need to ;-) ).

Just ask if you need any more help - and thanks!

jodh-intel:
Thanks for raising @knightXun. The message we are trying to convey by that warning is:

> You should only run these commands on a non-critical developer or test system.

You could add those words to the section if you want. The concern is that if we make this a "hint" (which is by definition optional), nobody will bother to read it. A warning might sound scary, but it's supposed to get your attention. Changing the warning to "Note" would work, but doesn't really convey "*you must read this*" to me.

What do others think? Maybe @klynnrif has some suggestions here on best practice?

Related: https://github.com/kata-containers/documentation/blob/master/Documentation-Requirements.md

klynnrif:
@knightXun / @grahamwhaley / @jodh-intel 
In this case I suggest keeping the "warning" label if things can get screwed up when this instruction isn't followed. If that was not the case, I think a "note" or "hint" would be appropriate. Thanks!

grahamwhaley:
@knightXun - it feels like we are leaning towards 'being safe' and leaving this as **Warning**, to ensure users and developers realise that running these command can/will/might/should modify their system.
wdyt? Happy if you can find others in Kata community to add their thoughts as well..

jodh-intel:
Hi @knightXun - Thanks again for raising, but as shown, we do think it is reasonable to keep the wording as "warning" as it should be "mildly scary" as we want everyone to read that section carefully :fearful: 

Are you happy for us to close this PR?

knightXun:
@jodh-intel ok, I agree with you, close it

Issue #458, opened 2019-05-08
doc requirements doc updates


jodh-intel:
Add requirement to add a TOC (table of contents) and also offer some advice on linking.

PR #459, opened 2019-05-08
Improve doc requirements

jodh-intel:
- Add a new "Linking advice" section to the documentation requirements document.
- State that all documents should contain a table of contents.

Fixes: #458.

### DIFF
--- a/Documentation-Requirements.md
+++ b/Documentation-Requirements.md
@@ -13,8 +13,10 @@ Containers](https://github.com/kata-containers) project.
 
 # General requirements
 
-- All documents are expected to be written in [GitHub Flavored Markdown](https://github.github.com/gfm) format.
-- All documents should have the `.md` file extension.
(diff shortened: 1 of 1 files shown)


Documentation-Requirements.md
```diff
@@ -20,6 +21,18 @@ All documents must:
 - Include a TOC (table of contents) at the top of the document with links to
   all heading sections.
 
+# Linking advice
+
+Linking between documents is strongly encouraged to help users and developers
+understand the material more easily. Linking also avoids repetition - if a
```
grahamwhaley: s/understand/navigate/ ? just a thought

jodh-intel: I've updated it with your suggestion as that probably makes more sense to most folk. But in my mind it's both: the linking between docs makes it easier to comprehend what this project is all about ;)

PR #461, opened 2019-05-08
how-to: fix kubadm command

### LINKED ISSUE #460: how-to-use-k8s-with-cri-containerd-and-kata.md is out of date
I run the kubeadm command (https://github.com/kata-containers/documentation/blob/master/how-to/how-to-use-k8s-with-cri-containerd-and-kata.md#start-kubernetes) but it fails and shows the following error:

```
Use "kubeadm init [command] --help" for more information about a command.
error: unknown flag: --skip-preflight-checks
```

devimc:
remove `--skip-preflight-checks` option since it has been deprecated

fixes #460

Signed-off-by: Julio Montes <julio.montes@intel.com>

### DIFF
--- a/how-to/how-to-use-k8s-with-cri-containerd-and-kata.md
+++ b/how-to/how-to-use-k8s-with-cri-containerd-and-kata.md
@@ -119,7 +119,7 @@ $ sudo systemctl daemon-reload
 - Start cluster using `kubeadm`
 
   ```bash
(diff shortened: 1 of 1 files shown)


chavafg:
@devimc what about removing that option? I think it is no longer needed

devimc:
@chavafg do you want me to remove it?

chavafg:
I think we originally have it because it was trying to look for docker socket, even when we specified another runtime, but that has been fixed. In addition there could be other errors that the user want to be aware of, so yeah, I'll prefer to not have it.
@egernst wdyt?

devimc:
@chavafg change applied, thanks

Issue #468, opened 2019-05-14
virtio-fs: intial how-to document


ganeshmaharaj:
Add the basic how-to for virtio-fs once the feature lands in master.

PR #469, opened 2019-05-14
how-to: virtio-fs initial documentation

ganeshmaharaj:
virtio-fs has landed as an experimental feature in kata. This patch
enable the basic how-to for this feature.

Fixes: #468
Signed-off-by: Ganesh Maharaj Mahalingam <ganesh.mahalingam@intel.com>

### DIFF
--- a/how-to/how-to-use-virtio-fs-with-kata.md
+++ b/how-to/how-to-use-virtio-fs-with-kata.md
@@ -1,13 +1,19 @@
-
 # Kata Containers with virtio-fs
 
-* [Introduction](#introduction)
-* [Pre-requisites](#pre-requisites)
-* [Configure Kata Containers](#configure-kata-containers)
+- [Introduction](#introduction)
+- [Pre-requisites](#pre-requisites)
+- [Install Kata Containers with virtio-fs support](#install-kata-containers-with-virtio-fs-support)
+- [Run a Kata Container utilizing virtio-fs](#run-a-kata-container-utilizing-virtio-fs)
 
 ## Introduction
 
-[virtio-fs](https://virtio-fs.gitlab.io/) is a shared file system that lets virtual machines access a directory tree on the host. In Kata, virtio-fs can be used to share the rootfs of the container on the host within the guest to provide significant performance improvements over 9p (the current default sharing mechanism).
+Container deployments utilize explicit or implicit file sharing between host filesystem and containers. From a trust perspective, avoiding a shared file-system between the trusted host and untrusted container is recommended. This is not always feasible. In Kata Containers, block-based volumes are prefered as they allow usage of either device pass through or virtio-blk for access within the virtual machine.
+
+As of the 1.7 release of Kata Containers, [9pfs](https://www.kernel.org/doc/Documentation/filesystems/9p.txt) is the default filesystem sharing mechanism. While this does allow for workload compatability, it does so with degraded performance and potential for POSIX compliance limitations.
+
+To help address these limitations, [virtio-fs](https://virtio-fs.gitlab.io/) has been developed. virtio-fs is a shared file system that lets virtual machines access a directory tree on the host. In Kata Containers, virtio-fs can be used to share container volumes, secrets, config-maps, configuration files (hostname, hosts, resolv.conf) and the container rootfs on the host with the guest.  virtio-fs provides significant performance and POSIX compliance improvements compared to 9pfs.
+
+Enabling of virtio-fs requires changes in the guest kernel as well as the VMM. For Kata Containers, experimental virtio-fs support is enabled through the [NEMU VMM](https://github.com/intel/nemu).
 
 **Note: virtio-fs support is experimental in the 1.7 release of Kata Containers. Work is underway to improve stability, performance and upstream integration. This is available for early preview - use at your own risk**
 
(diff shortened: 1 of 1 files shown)


how-to/how-to-use-virtio-fs-with-kata.md
```diff
@@ -0,0 +1,77 @@
...
+ #   - virtio-fs
+-shared_fs = "virtio-9p"
++shared_fs = "virtio-fs"
+
+ # Path to vhost-user-fs daemon.
+-virtio_fs_daemon = ""
++virtio_fs_daemon = "/home/foo/build-x86_64_virt/virtiofsd-x86_64"
+```
+
+As you can see from the snippet above, all you need to change are:
```
jodh-intel: I like the diff and the explanation. But the ideal would be to make this document ["executable"](https://github.com/kata-containers/tests/blob/master/.ci/kata-doc-to-script.sh) so that we can run it in the CI, to ensure it is always correct.

Doing that would require:

- Telling the user to install the `crudini` package.
- Calling `crudini` a few times to set the appropriate config values, like:
  ```bash
  $ file=/usr/share/defaults/kata-containers/configuration.toml
  $ sudo crudini --set "$file" hypervisor.qemu path '"/home/foo/build-x86_64_virt/x86_64_virt-softmmu/qemu-system-x86_64_virt"

wdyt? Can we arrange for the appropriate pieces to be built in the CI under `/home/foo/`?

how-to/how-to-use-virtio-fs-with-kata.md
```diff
@@ -0,0 +1,77 @@
...
+- The `shared_fs` option to `virtio_fs`.
+- The path of the virtiofsd daemon. `/home/foo/build-x86_64_virt/virtiofsd-x86_64` in this example.
+
+Once you have saved these modifications, you can start a new container:
+```bash
+$ docker run --runtime=kata-runtime -it busybox
+```
+And you will be able to verify this new container is running with the NEMU hypervisor and also using virtiofsd by looking for the hypervisor path and the virtiofs daemon process on the host.
+```bash
+$ ps -aux | grep virt
```
jodh-intel: Alternatively, you could run `kata-runtime kata-env`.

jodh-intel:
Hi @klynnrif - ptal.

how-to/how-to-use-virtio-fs-with-kata.md
ganeshmaharaj: i mirrored this part of the document similar to how the nemu document was written. Not sure how that works with cruidini today. @chavafg @GabyCT  any thoughts?

how-to/how-to-use-virtio-fs-with-kata.md
ganeshmaharaj: I wasn't sure that command will tell me about all the processes running on the host. we want to show the user that they will see the virtio-fs daemon running.

how-to/how-to-use-virtio-fs-with-kata.md
```diff
@@ -0,0 +1,77 @@
...
+**Note: This feature is still experimental and work is being done to get this production ready. Use at your own risk**
+
+This document describes how to get Kata Containers to work with virtio-fs.
+
+## Pre-requisites
+
+This document requires Kata Containers to be [installed](https://github.com/kata-containers/documentation/blob/master/install/README.md) on your system.
+
+* virtio-fs is currently only available with [NEMU](https://github.com/kata-containers/documentation/blob/master/how-to/how-to-use-kata-containers-with-nemu.md)
+* This feature currently requires the host to have hugepages support enabled. For eg: this can be enabled with `sysctl vm.nr_hugepages=1024` command on the host. `1024` is the default cache size as defined in kata's configuration file.
```
klynnrif: Suggested rewrite: 
This feature currently requires the host to have hugepages support enabled. Enable this with the `sysctl vm.nr_hugepages=1024` command on the host. ....

how-to/how-to-use-virtio-fs-with-kata.md
```diff
@@ -0,0 +1,77 @@
...
+
+## Configure Kata Containers
+
+All you need for this section is to modify the configuration file `/usr/share/defaults/kata-containers/configuration.toml` to specify the options related to this feature.
```
klynnrif: Suggested rewrite:
To configure Kata Containers, modify the configuration file `/usr/share/defaults/kata-containers/configuration.toml` to specify the options related to this feature.

how-to/how-to-use-virtio-fs-with-kata.md
```diff
@@ -0,0 +1,77 @@
...
+ #   - virtio-fs
+-shared_fs = "virtio-9p"
++shared_fs = "virtio-fs"
+
+ # Path to vhost-user-fs daemon.
+-virtio_fs_daemon = ""
++virtio_fs_daemon = "/home/foo/build-x86_64_virt/virtiofsd-x86_64"
+```
+
+As you can see from the snippet above, all you need to change are:
```
klynnrif: Suggested rewrite: 
As you can see from the previous snippet, you only need to change the following:

how-to/how-to-use-virtio-fs-with-kata.md
```diff
@@ -0,0 +1,77 @@
...
+- the path to the hypervisor binary. `/home/foo/build-x86_64_virt/x86_64_virt-softmmu/qemu-system-x86_64_virt` in this example.
+- The machine name from `pc` to `virt`,
+- The path of the firmware binary, `/usr/share/nemu/OVMF.fd` in this example,
+- The `shared_fs` option to `virtio_fs`,
+- The path of the virtiofsd daemon. `/home/foo/build-x86_64_virt/virtiofsd-x86_64` in this example.
+
+Once you have saved these modifications, you can start a new container:
```
klynnrif: Suggested rewrite: 
Once you save these modifications, start a new container:

how-to/how-to-use-virtio-fs-with-kata.md
```diff
@@ -0,0 +1,77 @@
...
+```bash
+$ docker run --runtime=kata-runtime -it busybox
+```
+Verify this new container is running with the NEMU hypervisor and also using virtiofsd by looking for the hypervisor path and the virtiofs daemon process on the host:
```
klynnrif: Suggested rewrite: 
Verify the new container is running with the NEMU hypervisor as well as using virtiofsd. To do this look for the hypervisor path and the virtiofs daemon process on the host:

how-to/how-to-use-virtio-fs-with-kata.md
jodh-intel: Ack.

how-to/how-to-use-virtio-fs-with-kata.md
jodh-intel: Understood. As doc is so critical though, I think we should ensure that as much of it as possible is "testable" as bitrotted documentation is useless to users. I'm happy to update the nemu doc with `crudini` if we think we can make both docs testable, to ensure their ongoing accuracy.

jodh-intel: I've raised #472 for the nemu doc updates...

how-to/how-to-use-virtio-fs-with-kata.md
```diff
@@ -0,0 +1,77 @@
...
+
+* [Introduction](#introduction)
+* [Pre-requisites](#pre-requisites)
+* [Configure Kata Containers](#configure-kata-containers)
+
+## Introduction
+
+[virtio-fs](https://virtio-fs.gitlab.io/) is a shared file system that lets virtual machines access a directory tree on the host. In Kata, virtio-fs can be used to share the rootfs of the container on the host within the guest to provide significant performance improvements over 9p (the current default sharing mechanism).
+
+**Note: This feature is still experimental and work is being done to get this production ready. Use at your own risk**
```
egernst:
```suggestion
**Note: virtio-fs support is experimental in the 1.7 release of Kata Containers. Work is underway to improve stability, performance and upstream integration. This is available for early preview - use at your own risk **
```

ganeshmaharaj:
@egernst that would be good to add as `you could also use kata-deploy like such....` and still provide the manual steps so they know what the bits are.

how-to/how-to-use-virtio-fs-with-kata.md
ganeshmaharaj: Wondering if we can do it as a follow-up change.

how-to/how-to-use-virtio-fs-with-kata.md
```diff
@@ -7,71 +6,43 @@
...
 
 **Note: virtio-fs support is experimental in the 1.7 release of Kata Containers. Work is underway to improve stability, performance and upstream integration. This is available for early preview - use at your own risk**
 
 This document describes how to get Kata Containers to work with virtio-fs.
 
 ## Pre-requisites
 
-This document requires Kata Containers to be [installed](https://github.com/kata-containers/documentation/blob/master/install/README.md) on your system.
-
-* virtio-fs is currently only available with [NEMU](https://github.com/kata-containers/documentation/blob/master/how-to/how-to-use-kata-containers-with-nemu.md)
```
ganeshmaharaj: Should we mention that it currently only works with NEMU here?

how-to/how-to-use-virtio-fs-with-kata.md
```diff
@@ -7,71 +6,43 @@
...
 * This feature currently requires the host to have hugepages support enabled. Enable this with the `sysctl vm.nr_hugepages=1024` command on the host.
 
-## Configure Kata Containers
```
ganeshmaharaj: Is the `virtio-fs the hard way` doc an enhancement for later?

egernst:
I pushed a commit.  Apologize for the toe stomp there.  I think the "how to do the hardway" isn't really ideal.  We actually package the nemu toml at this point, so the end user shouldn't have to go through and make modifications in any case.  The main part missing is a way to get nemu properly installed.

With those gaps present, I think just showing kata-deploy makes it straight forward (and, i think this is a friendly way for cloud-native folks to get started with Kata, anyway, imo).

how-to/how-to-use-virtio-fs-with-kata.md
egernst: If we want, it should be an enhancement imo.

how-to/how-to-use-virtio-fs-with-kata.md
egernst: Ah, I made more explicit.

how-to/how-to-use-virtio-fs-with-kata.md
```diff
@@ -0,0 +1,50 @@
...
+
+* This feature currently requires the host to have hugepages support enabled. Enable this with the `sysctl vm.nr_hugepages=1024` command on the host.
+
+## Install Kata Containers with virtio-fs support
+
+The Kata Containers NEMU configuration, the NEMU VMM and the virtiofs daemon are available in the [Kata Container release](https://github.com/kata-containers/runtime/releases) artifacts starting with the 1.7 release. While the feature is experimental, distribution packages are not supported, but installation is available through [kata-deploy](https://github.com/kata-containers/packaging/tree/master/kata-deploy).
+
+Install the latest release of Kata as follows:
+```
+docker run -v /opt/kata:/opt/kata -v /var/run/dbus:/var/run/dbus -v /run/systemd:/run/systemd -v /etc/docker:/etc/docker -it katadocker/kata-deploy kata-deploy-docker install
```
jcvenegas: OMG! nice! just in case I'd add `--runtime=runc` not a blocker

how-to/how-to-use-virtio-fs-with-kata.md
```diff
@@ -0,0 +1,50 @@
...
+## Running a Kata Container utilizing virtio-fs
+
+Once installed, start a new container, utilizing NEMU + virtiofs:
+```bash
+$ docker run --runtime=kata-nemu -it busybox
+```
+
+Verify the new container is running with the NEMU hypervisor as well as using virtiofsd. To do this look for the hypervisor path and the virtiofs daemon process on the host:
+```bash
+$ ps -aux | grep virt
```
jcvenegas: `virtiofsd`  could be better virt is a very generic prefix.

how-to/how-to-use-virtio-fs-with-kata.md
```diff
@@ -0,0 +1,77 @@
...
+**Note: virtio-fs support is experimental in the 1.7 release of Kata Containers. Work is underway to improve stability, performance and upstream integration. This is available for early preview - use at your own risk**
+
+This document describes how to get Kata Containers to work with virtio-fs.
+
+## Pre-requisites
+
+This document requires Kata Containers to be [installed](https://github.com/kata-containers/documentation/blob/master/install/README.md) on your system.
+
+* virtio-fs is currently only available with [NEMU](https://github.com/kata-containers/documentation/blob/master/how-to/how-to-use-kata-containers-with-nemu.md)
+* This feature currently requires the host to have hugepages support enabled. Enable this with the `sysctl vm.nr_hugepages=1024` command on the host.
```
stefanha: "enable_hugepages = true" is also required in /etc/kata-containers/configuration.toml.

jodh-intel: Thanks @stefanha - @ganeshmaharaj - could you raise a follow-up to fix this?

PR #474, opened 2019-05-16
Doc fixes

jodh-intel:
Lots of small doc fixes:

- Added missing doc links
- Removed stale list entries
- Fix partial sentence
- Add TOC
- Fix debian docker link


jodh-intel:
@egernst - I'm wondering if we want to merge "use-cases" and "how-to"'s...?

README.md
```diff
@@ -1,5 +1,15 @@
 # Documentation
 
+* [Getting Started](#getting-started)
```
grahamwhaley: ooi, do we have a 'standard' for where to put the toc? I believe I have posted some docs where the top lives above the first `#` top level heading for instance :-)

README.md
```diff
@@ -27,17 +27,16 @@ For details of the other Kata Containers repositories, see the
...
+### Howto guides
+
+See the [howto documentation](how-to).
 
 ## Kata Use-Cases
 
-* [VPP with Kata](./use-cases/using-vpp-and-kata.md)
-* [SR-IOV with Kata](./use-cases/using-SRIOV-and-kata.md)
 * [GPU Passthrough with Kata](./use-cases/GPU-passthrough-and-Kata.md)
+* [OpenStack Zun with Kata Containers](./use-cases/zun_kata.md)
```
grahamwhaley: I've also considered if use-cases and howtos should all be in one place - it's not really clear to me what the differentiation is, and sometimes I struggle to find docs due to the separation.

README.md
jodh-intel: We currently say ["at the top"](https://github.com/kata-containers/documentation/blob/master/Documentation-Requirements.md#general-requirements) but that can be before or after the optional top-level link which refers to the doc itself.

README.md
jodh-intel: Yes - the distinction is too subtle I think - let's just KISS and adopt one term consistently.

PR #478, opened 2019-05-16
docs: Advise use of non-interactive commands

### LINKED ISSUE #477: Suggest code blocks contain non-interactive commands
If a doc contains commands, they should be non-interactive where possible to allow for the possibility of automating the testing of the document in the CI.

jodh-intel:
If a doc contains commands, they should be non-interactive where possible to allow for the possibility of automating the testing of the document in the CI.

Fixes #477.

### DIFF
--- a/Documentation-Requirements.md
+++ b/Documentation-Requirements.md
@@ -28,6 +28,13 @@ All documents must:
 
   If you are adding a new document, ensure you add a link to it in the
   "closest" `README` above the directory where you created your document.
+- If the document needs to tell the user to manipulate files or commands, use a
+  [code block](#code-blocks) to specify the commands.
+
+  If at all possible, ensure that every command in the code blocks can be run
+  non-interactively. If this is possible, the document can be tested by the CI
+  which can then execute the commands specified to ensure the instructions are
+  correct. This avoids documents becoming out of date over time.
 
 # Linking advice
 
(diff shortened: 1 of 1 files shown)


Documentation-Requirements.md
```diff
@@ -28,6 +28,13 @@ All documents must:
 
   If you are adding a new document, ensure you add a link to it in the
   "closest" `README` above the directory where you created your document.
+- If the document needs to tell the user to manipulate files or commands, use a
+  [code block](#code-blocks) to specify the commands.
+
+  If at all possible, ensure that every command in the code blocks can be run
+  non-interactively. If this is possible, the document can be tested by the CI
+  which can then execute the commands specified to ensure the instructions are
+  correct. This avoids documents becoming out of date over time.
```
grahamwhaley: Maybe also mention that inter-spersing the resulting command output of the commands through the code block ... well, what should we do there? This is something I do fairly often. Hmm, maybe we can use `# comment blocks` to show the expected output without tripping up the CI doc code automation. For example,

```bash
$ ls
# output:
# fred.sh john.txt
$ ./fred.sh
```

iyswim?

jodh-intel: Your example will work fine in fact. Do you want me to add further details to the section that covers command output (https://github.com/kata-containers/documentation/blob/master/Documentation-Requirements.md#code-blocks)?

grahamwhaley: Please, whilst you are in there - I think if we have agreed a method, then we should stick it in the docs. thx

jodh-intel: The reason I'm asking is that that section already explains how to show command output, so the real question is do we want to add a second method for showing output?

jodh-intel:
Let's merge this as the existing doc already implicitly supports the scenario @grahamwhaley has described.

Issue #483, opened 2019-05-23
Add kata installation for all openSUSE distros


marcov:
Using `kata-manager.sh install-packages`, the result is:
- Tumbleweed: `ERROR: no install document for distro opensuse-tumbleweed`
- Leap >= 15.0: `ERROR: no install document for distro opensuse-leap`

PR #484, opened 2019-05-23
install: better support for all openSUSE distros

marcov:
Support installing kata packages on all openSUSE distros.
    
Fixes: #483

And also:
    install: improve Docker installation for openSUSE
    
    Editing systemd unit files to customize Docker for Kata may generate conflicts
    with what's specified in /etc/sysconfig/docker, so use that file directly.
    
    Also, libcgroup1 dependency is wrong for newer distros, and should be
    pulled automatically for older ones.

### DIFF
--- a/install/README.md
+++ b/install/README.md
@@ -49,7 +49,7 @@ Manual installation instructions are available for [these distributions](#suppor
 1. Add the Kata Containers repository to your distro package manager, and import the packages signing key.
 2. Install the Kata Containers packages.
 3. Install a supported container manager.
-4. Configure the container manager to use `kata-runtime` as the default OCI runtime. Or, for Kata Containers 1.5.0 or above, configure the 
+4. Configure the container manager to use `kata-runtime` as the default OCI runtime. Or, for Kata Containers 1.5.0 or above, configure the
    `io.containerd.kata.v2` to be the runtime shim (see [containerd runtime v2 (shim API)](https://github.com/containerd/containerd/tree/master/runtime/v2)
    and [How to use Kata Containers and CRI (containerd plugin) with Kubernetes](https://github.com/kata-containers/documentation/blob/master/how-to/how-to-use-k8s-with-cri-containerd-and-kata.md)).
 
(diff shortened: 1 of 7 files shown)


jodh-intel:
@marcov - it's worth pointing out that we don't currently have SUSE Jenkins slaves for this repo (only centos, ubuntu and fedora):

- http://jenkins.katacontainers.io/

... meaning that the CI won't be able to test these changes. I think it would make sense to add atleast openSUSE if we have capacity so I'd recommend raising an issue here: https://github.com/kata-containers/ci/issues/new.

marcov:
BTW, there's a problem with kata-doc-to-script and kata-manager.

The `ID` variable in `/etc/os-release` for all openSUSE's is set to `opensuse-leap` or `opensuse-tumbleweed`. 

I'd like to keep a singe file with instructions applicable for both, to avoid duplication, named like now `opensuse-installation-guide.md`.

I am not sure how to handle this. In kata-manager script we could add an exception to just use `opensuse` instead of the full `$ID`, but adding the same exception [in kata-doc-to-script install guide](https://github.com/kata-containers/documentation/blob/master/install/installing-with-kata-doc-to-script.md) would look weird.
It would be interesting to have something working like symlinks :)

marcov:
> BTW, there's a problem with kata-doc-to-script and kata-manager.
> ... [ CUT ] ...

My solution to the "problem" described above is to provide two extra files: `opensuse-tumbleweed-installation-guide` and `opensuse-leap-installation-guide.md` . They both have a link to the generic `opensuse-installation-guide.md`. 
When running `kata-doc-to-script` on these files, their (hidden) bash block is extracted, that prompts to run `kata-doc-to-script` again on the generic openSUSE guide. The script generated just contain an `echo` that prints:
```
NOTE: this document is just a link to the generic openSUSE install guide located at:
https://raw.githubusercontent.com/kata-containers/documentation/master/install/opensuse-installation-guide.md

Please download this file and run kata-doc-to-script.sh again.
```

marcov:
Stuck with the "all documents referenced" now :thinking:

jodh-intel:
Hi @marcov - yes, Travis is failing with this error:

```
ERROR: Document install/docker/opensuse-leap-docker-install.md is not referenced
```

That's due to https://github.com/kata-containers/tests/pull/1612. The error is trying to tell you that the document mentioned needs to be linked to from another `README` in this repo. In fact, all document (aside from a few like the top-level `README.md` itself) now need to be referenced. See:

- https://github.com/kata-containers/documentation/blob/master/Documentation-Requirements.md#general-requirements

Once you add links to the new doc you've added on this PR for leap and tumbleweed to:

- https://github.com/kata-containers/documentation/tree/master/install#supported-distributions

... I think Travis will be happy :smile:

jodh-intel:
Oh - you'll have to link to the docker install docs from the leap and tumbleweed docs too. Note that the markdown parsers cannot "see" the links hidden in HTML `<!-- ...-->` comments. If that's going to be a problem, we could always create a minimal `install/docker/README.md` that shows a table of docker install docs along with a suggestion that users probably want to start by looking at `install/README.md`.

marcov:
Thanks @jodh-intel. About the HTML `<!-- -->` comment, this is the reason:

Install instructions for openSUSE can be made common for all versions, so I preferred keeping a single file with the instructions, and have "placeholder" files named after each versions that point to it.

When following [kata-doc-to-script install guide](https://github.com/kata-containers/documentation/blob/master/install/installing-with-kata-doc-to-script.md), you can't easily point to generic instruction file without bloating the bash code block to copy/paste.

So, following that guide e.g. on openSUSE Leap and running the generated script gives:
```bash
# ./opensuse-leap-install.sh 
NOTE: this document is just a link to the generic openSUSE install guide located at:
https://raw.githubusercontent.com/kata-containers/documentation/master/install/opensuse-installation-guide.md

Please download this file and run kata-doc-to-script.sh again.
```

jodh-intel:
@marcov - nice!

jodh-intel:
Just one more ack needed over here folks.

Issue #489, opened 2019-06-08
error in vCPU calculation documentation


egernst:
See https://github.com/kata-containers/documentation/blob/master/design/cpu-constraints.md

The formula is listed as:
```
(quota + (period -1)) / period
```

This doesn't match the code, and it is not correct.  vCPU = quota / period

This is related to https://github.com/kata-containers/runtime/issues/1767

shsha-msft:
Just to clarify. Is the following understanding correct?

For containers in a POD the formula is vCPU = quota / period
For the POD itself the formula is : (quota + (period -1)) / period

/cc @gkhanna79

jcvenegas:
The calculation is as follow:

The containers in a pod are converted to milli cpus
```go
func CalculateMilliCPUs(quota int64, period uint64) uint32 {

        // If quota is -1, it means the CPU resource request is
        // unconstrained.  In that case, we don't currently assign
        // additional CPUs.
        if quota >= 0 && period != 0 {
                return uint32((uint64(quota) * 1000) / period)
        }

        return 0
}
```

Example:
``` go
//  This is 1500
milliCPUs:= CalculateMilliCPUs(150000, 100000)
```

When a pod has mutiples containers the milliCPUs are sum:
``` go
millCPUs := 0
milliCPUs += CalculateMilliCPUs(150000, 100000)
milliCPUs += CalculateMilliCPUs(150000, 100000)
// This result in 3000 millcpus
```
After the milliCPUs  for a POD are calculated this is converted to CPUs with:
```
//CalculateVCpusFromMilliCpus converts from mCPU to CPU, taking the ceiling
// value when necessary
func CalculateVCpusFromMilliCpus(mCPU uint32) uint32 {
       return (mCPU + 999) / 1000
}
```
VCPUS:
(3000 + 999 ) / 1000 = 3

egernst:
@jcvenegas overall, though, the documentation is wrong.  

It should be  vCPU = ceiling( quota / period )

Can you send a PR to update the doc?  WDYT?

PR #495, opened 2019-06-11
docs: cpu: resources: fix vcpu calculation formula

jcvenegas:
The formula is not updated according on
how is done in kata-runtime.

Fixes: #489

Signed-off-by: Jose Carlos Venegas Munoz <jose.carlos.venegas.munoz@intel.com>


egernst:
I would actually clarify in the commit message that this is a fix; it isn't so much that it doesn't match . the code, its that it was wrong?

Issue #507, opened 2019-06-19
docs: Use 4.19.x config file when building kernel


nitkon:
Update Developer Guide steps to use 4.19.x kernel config file.

PR #508, opened 2019-06-19
docs: Use kernel build script for building kernel in developer guide

nitkon:
Use 4.19.x config file when building kernel. 

Fixes: #507

### DIFF
--- a/Developer-Guide.md
+++ b/Developer-Guide.md
@@ -330,40 +330,7 @@ $ (cd /usr/share/kata-containers && sudo ln -sf "$image" kata-containers-initrd.
 
 # Install guest kernel images
 
-As a prerequisite, you need to install `libelf-dev` and `bc`. Otherwise, you
-will not be able to build the kernel from sources.
-
-```
-$ go get github.com/kata-containers/tests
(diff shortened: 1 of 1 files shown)


grahamwhaley:
Agree. I already asked/suggested we do that pretty recently when somebody else PRd the same or similar change I think - we already have the kernel build docs and script over at https://github.com/kata-containers/packaging/tree/master/kernel including documentation. Can you ammend this PR to point to those instructions instead please @nitkon , then we won't have to play catchup here every time we make a kernel version change.

nitkon:
@jodh-intel @grahamwhaley : PTAL ...

nitkon:
@grahamwhaley : Done. 👍

eadamsintel:
I just tried build-kernel.sh on Clear Linux and it complained that the arch command was missing which happened previously with OSbuilder. This should be easily replaceable with a uname command instead.

`
go/src/github.com/kata-containers/tests/.ci/lib.sh: line 38: arch: command not found
`

It seemed to build the kernel OK and I was able to boot up a kata container with that kernel.

PR #519, opened 2019-07-24
architecture.md: Fix networking info

nitkon:
The networking part of the doc talks about the 
two ends of the veth pair. One end is in the container
networking namespace and the other one should
be in the host networking namespace. Fix this info.

Fixes: #518


amshinde:
@grahamwhaley This does things clearer.

nitkon:
All green! @GabyCT @devimc

Issue #526, opened 2019-08-01
docs: minikube: note that kata-fc does not presently work under minikube


grahamwhaley:
`kata-fc` fails to run under `minikube` at present, due to lack of (at least) block based storage volumes. Not that in the install docs.

PR #527, opened 2019-08-01
minikube: note kata-fc does not function under minikube

grahamwhaley:
`kata-fc` does not presently function under `minikube` due to
lack of block based storage. Make that clear in the installation
documents, to help prevent users going through the whole install
process, only to be disappointed when they find ti does not work.

Fixes: #526

### DIFF
--- a/install/minikube-installation-guide.md
+++ b/install/minikube-installation-guide.md
@@ -32,6 +32,9 @@ This installation guide has only been verified under a Minikube Linux installati
 > - This installation guide may not work for macOS installations of Minikube, due to the lack of
 nested virtualization support on that platform.
 > - This installation guide has not been tested on a Windows installation.
+> - Kata under Minikube does not currently support Kata Firecracker (`kata-fc`).
+>   Although the `kata-fc` binary will be installed as part of these instructions,
+>   via `kata-deploy`, pods cannot be launched with `kata-fc`, and will fail to start.
 
 Before commencing installation, it is strongly recommended you read the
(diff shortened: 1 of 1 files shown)


install/minikube-installation-guide.md
```diff
@@ -32,6 +32,10 @@ This installation guide has only been verified under a Minikube Linux installati
 > - This installation guide may not work for macOS installations of Minikube, due to the lack of
 nested virtualization support on that platform.
 > - This installation guide has not been tested on a Windows installation.
+> - Kata under Minikube does not currently support Kata Firecracker (`kata-fc`), due to
+>   lack of a block based storage driver availability. Although the `kata-fc` binary will be
```
jodh-intel: Nit: I think the word "availability" is redundant?

> due to the lack of a block based storage driver.

grahamwhaley: hmm, given the thread that spawned this update, I quite like the way it confirms why there is a lack of driver... implies that it is not something one can just go update/fix trivially. I'll change it if you really want though...?

jodh-intel: It just doesn't read well to me, but I'm struggling to think how to improve the wording. I'll ack and see if @klynnrif has any thoughts...

grahamwhaley: np. I simplified and dropped the whole section about why it will not work, and now just note it won't work... (and, the reasons why may change or be incomplete, so just saying 'does not work' should have a longer shelf life, until the day it does work...)

Issue #542, opened 2019-08-27
docs: Add documentation about host cgroup management


jcvenegas:
Document how cgroups are done today and what is expected
for the upcoming SandboxCgroupOnly option.

PR #543, opened 2019-08-27
docs: Add documentation about host cgroup management

jcvenegas:
Document how cgroups are done today and what is expected
for the upcoming SandboxCgroupOnly option.

Prior cgroup documentation are no longer accurate. Removing the cgroup
discussion from the cpu sizing discussion. Updating the
cpu-constraints.md file name to reflect this.

Fixes: #542

Signed-off-by: Eric Ernst <eric.ernst@intel.com>
Signed-off-by: Jose Carlos Venegas Munoz <jose.carlos.venegas.munoz@intel.com>

### DIFF
--- a/design/host-cgroups.md
+++ b/design/host-cgroups.md
@@ -1,0 +1,208 @@
+- [Host cgroup management](#host-cgroup-management)
+  - [Introduction](#introduction)
+  - [`SandboxCgroupOnly` enabled](#sandboxcgrouponly-enabled)
+    - [What does Kata do in this configuration?](#what-does-kata-do-in-this-configuration)
+    - [Why create a Kata-cgroup under the parent cgroup?](#why-create-a-kata-cgroup-under-the-parent-cgroup)
+    - [Improvements](#improvements)
+  - [`SandboxCgroupOnly` disabled (default, legacy)](#sandboxcgrouponly-disabled-default-legacy)
+    - [What does this method do?](#what-does-this-method-do)
+      - [Impact](#impact)
+  - [Summary](#summary)
+
+# Host cgroup management
+
+## Introduction
+
+In Kata Containers, workloads run in a virtual machine that is managed by a virtual
+machine monitor (VMM) running on the host. As a result, Kata Containers run over two layers of cgroups. The
+first layer is in the guest where the workload is placed, while the second layer is on the host where the
+VMM and associated threads are running.
+
+The OCI [runtime specification][linux-config] provides guidance on where the container cgroups should be placed:
+
+  > [`cgroupsPath`][cgroupspath]: (string, OPTIONAL) path to the cgroups. It can be used to either control the cgroups
+  > hierarchy for containers or to run a new process in an existing container
+
+cgroups are hierarchical, and this can be seen with the following pod example:
+
+- Pod 1: `cgroupsPath=/kubepods/pod1`
+  - Container 1:
+`cgroupsPath=/kubepods/pod1/container1`
+  - Container 2:
+`cgroupsPath=/kubepods/pod1/container2`
+
+- Pod 2: `cgroupsPath=/kubepods/pod2`
+  - Container 1:
+`cgroupsPath=/kubepods/pod2/container2`
+  - Container 2:
+`cgroupsPath=/kubepods/pod2/container2`
+
+Depending on the upper-level orchestrator, the cgroup under which the pod is placed is
+managed by the orchestrator. In the case of Kubernetes, the pod-cgroup is created by Kubelet,
+while the container cgroups are to be handled by the runtime. Kubelet will size the pod-cgroup
+based on the container resource requirements.
+
+Kata Containers introduces a non-negligible overhead for running a sandbox (pod). Based on this, two scenarios are possible:
+ 1) The upper-layer orchestrator takes the overhead of running a sandbox into account when sizing the pod-cgroup, or
+ 2) Kata Containers do not fully constrain the VMM and associated processes, instead placing a subset of them outside of the pod-cgroup.
+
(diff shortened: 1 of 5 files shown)


egernst:
Hey Carlos, as a first step, I think we should probably combine _cpu-constraints.md_ with your _host-cgroups.md_, probably into a document named something like, _constraining-kata-on-host.md_.

jcvenegas:
@jodh-intel @egernst @chavafg @devimc thanks for the suggestions - document update.

design/host-cgroups.md
```diff
@@ -0,0 +1,223 @@
...
+
+#### Impact
+
+If resources are reserved at a system level to account for the overheads of
+running sandbox containers, this configuration can be utilized with adequate
+stability. In this scenario, non-negligible amounts of CPU and memory will be
+utilized unaccounted for on the host.
+
+[linux-config]: https://github.com/opencontainers/runtime-spec/blob/master/config-linux.md
+[cgroupspath]: https://github.com/opencontainers/runtime-spec/blob/master/config-linux.md#cgroups-path
```
jodh-intel: I think it would be useful if you could add a table summarising the different options and the pros and cons of both. Something like this (but with more columns ;)

| cgroup option | default? | legacy |
|-|-|-|
| `SandboxCgroupOnly` | yes | no |
| `SandboxCgroupOnly` | no | yes |

From the users perspective, it would be good if they could scan the table to determine:

- Which cgroup option is appropriate for them.
  - Depending on the orchestrator they are using.

Other columns / thoughts you might want to add / consider:

- host cgroup or container cgroup?
- memory density implication
- security implication / isolation level (if any)
- where the overhead exists
- whether the runtime limits host cgroup
- accounting / statistics impact

egernst:
@jodh-intel I updated the commit based on your valuable feedback. PTAL?

design/host-cgroups.md
```diff
@@ -0,0 +1,230 @@
...
+# Host cgroup management
+
+## Introduction
+
+Kata Containers runs over two layers of cgroups, the first layer is in the
+guest where only the workload is placed, the second layer is in the host and it
+will be explained on this document.
+
+```
+$ docker run -dt --runtime=kata-runtime nginx
```
jodh-intel: Could we split this block into something like:

> For example, when running this command:
> ```sh
> $ docker run -dt --runtime=kata-runtime nginx
> ```
>
> The following cgroups are created:
>
> << ASCII graphics awesomeness goes here >>

design/host-cgroups.md
```diff
@@ -0,0 +1,230 @@
...
+
+In Kata Containers, workloads run in a virtual machine that is managed by a virtual
+machine monitor (VMM) running on the host. As a result, Kata Containers run over two layers of cgroups. The
+first layer is in the guest where the workload is placed, while the second layer is on the host where the
+VMM and associated threads are running.
+
+The OCI [runtime specification][linux-config] provides guidance on where the container cgroups should be placed:
+
+- [`CgroupsPath`][cgroupspath]
+> `cgroupsPath` (string, OPTIONAL) path to the cgroups. It can be used to either control the cgroups
```
jodh-intel: This needs to be indented so it's a child of its parent bullet.

design/host-cgroups.md
```diff
@@ -0,0 +1,230 @@
...
+Depending on the upper-level orchestrator, the cgroup under which the pod is placed is
+managed by the orchestrator. In the case of Kubernetes, the pod-cgroup is created by Kubelet,
+while the container cgroups are to be handled by the runtime. Kubelet will size the pod-cgroup
+based on the container resource requirements.
+
+Kata Containers introduces a non-negligible overhead for running a sandbox (pod). Based on this, two scenarios are possible:
+ 1) The upper-layer orchestrator takes the overhead of running a sandbox into account when sizing the pod-cgroup, or
+ 2) Kata Containers do not fully constrain the VMM and associated processes, instead placing a subset of them outside of the pod-cgroup.
+
+Kata Containers provides two options for how cgroups are handled on the host.
```
jodh-intel: Obvious question that a newby might ask: "where are these settings? How do I enable them?" We should point to https://github.com/kata-containers/runtime#configuration.

design/host-cgroups.md
```diff
@@ -0,0 +1,230 @@
...
+
+[linux-config]: https://github.com/opencontainers/runtime-spec/blob/master/config-linux.md
+[cgroupspath]: https://github.com/opencontainers/runtime-spec/blob/master/config-linux.md#cgroups-path
+
+## Summary
+
+| cgroup option | default? | status | pros | cons
+|-|-|-|-|-|
+| `SandboxCgroupOnly=false` | yes | legacy | Easiest to make Kata work | Unaccounted for memory and resource utilization
+| `SandboxCgroupOnly=true` | no | recommended | Complete tracking of Kata memory and CPU utilization. In Kubernetes, the Kubelet can fully constrain Kata via the pod cgroup | Requires upper layer orchestrator which sizes sandbox cgroup appropriately |
```
jodh-intel: Nice! How would the user know (or determine) if their upper layer orchestrator sizes sandbox cgroups though? Can we give any further info such as "k8s vX.Y.Z or newer" / "Docker vX.Y.Z or newer" here?

design/host-cgroups.md
egernst: actually.... I want to just remove this part of the intro...

design/host-cgroups.md
egernst: I'd like to add this in a follow on PR, once the 1.16 Kubernetes documentation is published.  Then I can link to this (should be out in the next week or two).

egernst:
/test

PTAL - updated based on feedback

Issue #551, opened 2019-09-04
Clarify Docker configuration details


jodh-intel:
Improve the Docker installation instructions by making it clear *which* of the multiple ways of configuration Docker for Kata is the default, and that it is not necessary to do anything further if users select the automatic installation method.

Related: https://github.com/kata-containers/documentation/issues/550.

PR #552, opened 2019-09-04
docs: Emphasise default docker config

jodh-intel:
Improved the Docker installation instructions by making it clear *which* of the multiple ways of configuration Docker for Kata is the default, and that it is not necessary to do anything further if users select the automatic installation method.

Fixes: #551.


jodh-intel:
Yeah :)

/test

Issue #556, opened 2019-09-17
Does rootfs-builder support debian/ubuntu and suse distro ?


Kvasscn:
In the Developer-Guide document description(follow this link https://github.com/kata-containers/documentation/blob/master/Developer-Guide.md#create-a-rootfs-image), we can choose one of alpine, centos, clearlinux, euleros, and fedora distro,  but i'm not sure debian/ubuntu/suse whether or not supported, because the kata osbuilder packages already have these distro's build scripts.

PR #557, opened 2019-09-17
Developer-Guide.md: update rootfs build description

Kvasscn:
Add rootfs distro choices for debian, ubuntu, and suse.

Fixes: #556.

Signed-off-by: zhanghj.lc <zhanghj.lc@inspur.com>


Kvasscn:
@jodh-intel  yes, it's a better way, but the distro names in the compatibility matrix are not acceptable parameter for the rootfs.sh build script. for example, "CentOS" consists of upper and lower case letters,
and does not match the input parameter list of rootfs.sh.

PR #578, opened 2019-12-18
install: Update Fedora instructions

fidencio:
kata packages are already present on Fedora 31+. Knowing that, there's
no reason to use the packages provided by OpenSUSE.

Signed-off-by: Fabiano Fidêncio <fidencio@redhat.com>

### DIFF
--- a/install/fedora-installation-guide.md
+++ b/install/fedora-installation-guide.md
@@ -3,12 +3,7 @@
 1. Install the Kata Containers components with the following commands:
 
    ```bash
-   $ source /etc/os-release
-   $ ARCH=$(arch)
-   $ BRANCH="${BRANCH:-master}"
-   $ sudo dnf -y install dnf-plugins-core
-   $ sudo -E dnf config-manager --add-repo "http://download.opensuse.org/repositories/home:/katacontainers:/releases:/${ARCH}:/${BRANCH}/Fedora_${VERSION_ID}/home:katacontainers:releases:${ARCH}:${BRANCH}.repo"
-   $ sudo -E dnf -y install kata-runtime kata-proxy kata-shim
+   $ sudo dnf -y install kata-runtime
    ```
 
 2. Decide which container manager to use and select the corresponding link that follows:


devimc (approved):

fidencio:
Shall I create an issue to link in the patch?

devimc:
@fidencio yes, and include `fixes #xxx` in your commit message

install/fedora-installation-guide.md
```diff
@@ -2,6 +2,14 @@
 
 1. Install the Kata Containers components with the following commands:
 
+   1.1. From Fedora 31 onwards:
+
+   ```bash
+   $ sudo dnf -y install kata-runtime kata-proxy kata-shim
```
jodh-intel: afaik, you can just install the Fedora `kata-runtime` package which will auto-install the proxy and shim.

However, does that give you a working system? The OBS packages we produce also provide a container kernel and the disk image / initramfs files (`kata-linux-container` and `kata-containers-image` packages (see http://download.opensuse.org/repositories/home:/katacontainers:/releases:/x86_64:/master/Fedora_30/x86_64/)) but I'm not sure you get that with F31 (yet)?

/cc @jcvenegas, @lsm5.

install/fedora-installation-guide.md
```diff
@@ -2,6 +2,14 @@
...
+   ```
+
+   1.2. For Fedora 30 or older:
```
jodh-intel: Although these instructions are fine for humans, we actually "execute" these documents as scripts in the CI to ensure that they generate a working system.

As such, we need to explain in words that F31 and newer can use the Fedora packages. But we also need logic so that the CI can dtrt. Something like:

<pre>
```bash
$ source /etc/os-release
$ if [ $VERSION_ID -ge 31 ]; then
    sudo dnf -y install kata-runtime
else
  # the existing block of code
  ARCH=$(arch)
  ....
fi
```
</pre>

jodh-intel (commented):
Thanks for raising @fidencio. There are a couple of non-obvious issues here though.

jodh-intel:
Details on how we "execute" documents (purely for interests sake):

- https://github.com/kata-containers/documentation/blob/master/Documentation-Requirements.md#code-blocks
- https://github.com/kata-containers/tests/tree/master/cmd/kata-manager
- https://github.com/kata-containers/tests/blob/master/.ci/kata-doc-to-script.sh

install/fedora-installation-guide.md
fidencio: Aha! Okay, so, my proposal is to close this MR (and the issue associated) and get back to this one whenever Fedora has all the packages in place.

Does that work for everyone involved?

sameo: @c3d Isn't F31 shipping the a rootfs and kernel (Fedora based) as well?

fidencio: Interesting. When taking a look at https://copr.fedorainfracloud.org/coprs/ddd/kata/ I see the kata-containers-image. However, seems that @c3d has decided to focus on the other packages for Fedora 31.

crobinso: In Fedora, kata-runtime pulls in kata-osbuilder. When kata-osbuilder is installed, rpm %post runs on the host, and calls kata-osbuilder build scripts, which invoke dracut on the host which builds an 'image' from installed host content. Some custom packaging bits also cp current host kernel/initrd from /boot into /usr/share/kata-containers and symlinks them at the expected paths for configuration.toml to pick up.

The kata-linux-container/kata-containers-image way of doing things doesn't really work with Fedora packaging policies, so we went with this approach. It's not impossible in the future we would ship a stripped down Fedora kernel specifically for kata but it would be unprecedented in Fedora land.

fidencio:
Okay, based on @crobinso's explanation and the @jodh-intel's review, I've updated the patch.

jodh-intel:
Thanks @fidencio. We do still need to understand the situation regarding the kata kernel and image though. How are users expected to obtain these assets on F31+?

jodh-intel:
Doh! Looks like I didn't get a mail about the latest update there. I'd like to test this tomorrow before acking, but thanks for updating!

amshinde:
@jodh-intel I had tried kata on fedora 31 the other day, and it does not work out of the box.
First, I realised selinux needed to be disabled. We need to brainstorm what needs to be done on Kata side to add this support.
Fedora 31 also ships with cgroupsv2 support. We are currently working on adding that. 
So, one needs to explicitly fallback to cgroupsv1 for Kata.

Booting from rootfs fails as kernel lacks NVDIMM support.  So today just initrd works.

Complete list of bugs, some of them non-blocking:
https://apps.fedoraproject.org/packages/kata-runtime/bugs

crobinso:
Yes I should have mentioned, the 'out of the box' piece has some problems. What we have been telling people is:

```console
# onetime setup per host boot
sudo mkdir -p /run/cgroupv1; sudo mount -t cgroup -o cpu,cpuacct cgroup /run/cgroupv1/

# Run a sample container
sudo podman run --security-opt label=disable --runtime=/usr/bin/kata-runtime --rm -it alpine sh 
```

First part will set up enough of cgroupv1 to make kata happy (last I checked). You can also boot the host with `systemd.unified_cgroup_hierarchy=0`

The podman `--security-opt label=disable` removes the need to fully disable selinux.
podman also provides a helpful `--runtime` option that saves having to make system defaults changes.
FWIW on Fedora docker does not get a lot of love

The nvdimm/image boot issue is known, due to the driver not being built into the kernel. Unfortunately we are at an impasse on that one at the moment: https://bugzilla.redhat.com/show_bug.cgi?id=1750581

amshinde:
@crobinso Thanks, passing `--security-opt label=disable` works too. I had tried with disabling selinux on a system level earlier. With the above steps you mentioned I got kata working with podman.

We should add the above steps to the documentation in this PR.

Since cgroupsv2 is the default, docker does not work either out of box on fedora31. The only way I got it working is by passing  `systemd.unified_cgroup_hierarchy=0` on the kernel command line and rebooting. @crobinso Is that the recommended way to go about for enabling docker ?

Lastly, I noticed that the configuration.toml file for kata exists under `/usr/share/kata-containers/defaults/configuration.toml` in fedora31 instead of `/usr/share/defaults/kata-containers/configuration.toml`. Was there any particular reason for picking the former.
This needs a mention in our docs as well.

c3d:
> On 19 Dec 2019, at 02:15, Archana Shinde <notifications@github.com> wrote:
> 
> @crobinso Thanks, passing --security-opt label=disable works too. I had tried with disabling selinux on a system level earlier. With the above steps you mentioned I got kata working with podman.
> 
> We should add the above steps to the documentation in this PR.
> 
> Since cgroupsv2 is the default, docker does not work either out of box on fedora31. The only way I got it working is by passing systemd.unified_cgroup_hierarchy=0 on the kernel command line and rebooting. @crobinso Is that the recommended way to go about for enabling docker ?

I have other issues with docker so I have
> 
> Lastly, I noticed that the configuration.toml file for kata exists under /usr/share/kata-containers/defaults/configuration.toml in fedora31 instead of /usr/share/defaults/kata-containers/configuration.toml. Was there any particular reason for picking the former.

Not many packages seem to put things directly in /usr/share/defaults on Fedora. Any reason why it matters for Kata (besides instructions)?

jodh-intel:
Adding dnm label whilst we work through these issues...

GabyCT:
@jodh-intel @fidencio what is the status of this instructions? should we add the specification that to run docker with fedora 31 we need cgroupsv1?

jodh-intel:
imho our install docs / scripts cannot reasonably modify the way a users *entire system* boots and operates just to accommodate Docker.

Working on that not unreasonable assumption, I think the best we can do is to ensure the install docs detect cgroups v2 and error with a helpful message that:

- Explains the cgroups v1/v2 issue.
- States that installing (docker) on distro versions using cgroups v2 (such as F31) require manual intervention.
- Listing what intervention is required along with pros / cons.

In fact, I think this topic should be discussed by the @kata-containers/architecture-committee.

install/fedora-installation-guide.md
```diff
@@ -3,12 +3,7 @@
 1. Install the Kata Containers components with the following commands:
 
    ```bash
-   $ source /etc/os-release
-   $ ARCH=$(arch)
-   $ BRANCH="${BRANCH:-master}"
-   $ sudo dnf -y install dnf-plugins-core
-   $ sudo -E dnf config-manager --add-repo "http://download.opensuse.org/repositories/home:/katacontainers:/releases:/${ARCH}:/${BRANCH}/Fedora_${VERSION_ID}/home:katacontainers:releases:${ARCH}:${BRANCH}.repo"
-   $ sudo -E dnf -y install kata-runtime kata-proxy kata-shim
+   $ sudo dnf -y install kata-runtime
```
c3d, line 6: I think that it would be nice to explain that we have both `kata-runtime` that installs 1.x and `kata-containers` that installs 2.x. That would probably avoid confusion about what that `kata-containers` thing is.

c3d (changes requested):
Suggestion: mention the difference between 1.x and 2.0 packages.

install/fedora-installation-guide.md
fidencio: Hmm.  I'm not sure here.  This repo is specific about 1.x.  Mentioning `kata-containers` and 2.x only for Fedora seems inconsistent.  Changing the whole documentation to mention this about every distro / snap / etc feels like not worth it.

I'd appreciate if I could get the ACK and merge it as it's.

GabyCT (approved):

fidencio:
Oh, this one was superseded by https://github.com/kata-containers/documentation/pull/711.  Good to know! :-)

Thanks a lot for the reviews, I'm closing this one.

Issue #584, opened 2020-01-20
Add Debian 10 to list of available distros


jodh-intel:
Updated the list of supported distros now we create packages for Debian 10.

drcursor:
Even though its stated as being supported, I can't find the packages under http://download.opensuse.org/repositories/home:/katacontainers:/releases:/x86_64:/master/

PR #577, opened 2019-12-18
Add Debian 10

mnaser:
It seems like we're building packages for Debian 10 right now anyways.

PR #585, opened 2020-01-20
docs: Add Debian 10 to distro table

jodh-intel:
Now that [1] has landed, update the list of supported distros to include
Debian 10.

[1] - https://github.com/kata-containers/packaging/issues/647

Fixes: #584.

### DIFF
--- a/install/README.md
+++ b/install/README.md
@@ -42,7 +42,7 @@ Kata is packaged by the Kata community for:
 |Distribution (link to installation guide)                        | Versions                                                                                                          |
 |-----------------------------------------------------------------|-------------------------------------------------------------------------------------------------------------------|
 |[CentOS](centos-installation-guide.md)                           | 7                                                                                                                 |
-|[Debian](debian-installation-guide.md)                           | 9                                                                                                                 |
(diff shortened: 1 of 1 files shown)


jodh-intel:
Replaces #577 which hasn't been updated in a while.

grahamwhaley:
CI spell check fail @jodh-intel - I'm sort of hoping it was not due to https://github.com/kata-containers/tests/pull/2228
CI logs are loooong though.... any thoughts?

jodh-intel:
@grahamwhaley - Ufortunately, I think it is failing due to your PR, yes.

Since that script runs with `set -e` and since none of the spell checks failed the function is ending with:

```sh
[ $docs_failed -ne 0 ] && die ...
```

which expands to:

```sh
[ 0 -ne 0 ] && die ...
```

which expands to:

```sh
false && die ...
```

So since that function is returning `false` (`1`), the script just aborts at that point. If you change that line to the following I think it will work:

```sh
[ $docs_failed -ne 0 ] && die "..." || true
```

grahamwhaley:
ack. see https://github.com/kata-containers/tests/pull/2240
I wonder, can we test that with this PR with a `depends on` ??

Issue #593, opened 2020-02-13
config: debug: how to capture kernel boot messages


grahamwhaley:
**Which feature do you think can be improved?**

We don't document the minimum config chances you need to make to capture the kernel boot messages from the VM. Sometimes this is handy (if you are developing new config options for instance, and your containers are not booting), and you don't need to enable full debug to do this.

**How can it be improved?**

Document the minimum setup required, and how to view the messages.

PR #594, opened 2020-02-13
Developer: document how to get kernel boot messages

grahamwhaley:
Add a section detailing the minimum debug you need to configure in
order to capture the kernel boot messages in the system journal.

Fixes: #593

### DIFF
--- a/Developer-Guide.md
+++ b/Developer-Guide.md
@@ -44,6 +44,7 @@
         * [Create a container](#create-a-container)
         * [Connect to the virtual machine using the debug console](#connect-to-the-virtual-machine-using-the-debug-console)
         * [Obtain details of the image](#obtain-details-of-the-image)
+    * [Capturing kernel boot logs](#capturing-kernel-boot-logs)
     * [Running standalone](#running-standalone)
         * [Create an OCI bundle](#create-an-oci-bundle)
         * [Launch the runtime to create a container](#launch-the-runtime-to-create-a-container)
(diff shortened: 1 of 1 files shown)


grahamwhaley:
@devimc @dagrh - thought you might review.
@devimc - this may only work in non-vsock mode or?? Can you elaborate and let me know if I can/should/need to improve it... thx.

Developer-Guide.md
```diff
@@ -569,6 +570,52 @@ file exists and contains details of the image and how it was created:
...
+a shell prompt inside the container to view the kernel boot logs.
+
+If however you are unable to `exec` into the container, you can enable some debug
+options to have the kernel boot messages logged into the system journal.
+
+To achieve this you need to `enable_debug=true` in the Kata Configuration file
+for both the hypervisor/kernel and the proxy configuration sections. The following
+shows the differences as required in the `configuration-qemu.toml` file for instance:
+
+```patch
```
devimc: oii, I wouldn't like to include a path here, because configuration files tends to change, hence the patch context may not match causing hunks . Keeping this patch up-to-date can be tedious and painful.
There is already a section in this file that mentions how to enable debugging mode [1], how about we mention it instead of this patch?

[1] - https://github.com/kata-containers/documentation/blob/master/Developer-Guide.md#enable-full-debug

Developer-Guide.md
```diff
@@ -569,6 +570,52 @@ file exists and contains details of the image and how it was created:
...
++enable_debug = true
+
+ [shim.@PROJECT_TYPE@]
+ path = "@SHIMPATH@"
+```
+
+The kernel boot messages will appear in the `kata-proxy` log such as:
+
+```bash
+$ journalctl -t kata-proxy
```
devimc: in some systems `sudo` is required

Developer-Guide.md
grahamwhaley: ack. Reworked to show the two different sets of debug that might need to be enabled.

Developer-Guide.md
grahamwhaley: ack. Originally I tried to prefix with a root `#` prompt, but not only might that be too subtle, but it rendered as a comment! - updated.

devimc:
thanks @grahamwhaley

lifupan:
Hi @grahamwhaley , I think this instruction is only worked with cli mode, and not the shimv2.  Do you think is it better to append the shimv2 case here or I make another PR to address it after this PR merged.

Developer-Guide.md
```diff
@@ -569,6 +570,41 @@ file exists and contains details of the image and how it was created:
...
+Which debug options you enable depends on if you are using the hypervisor `vsock` mode
+or not, as defined by the `use_vsock` setting in the `[hypervisor.qemu]` section of
+the configuration file. The following details the settings:
+
+- For `use_vsock = false`:
+    - Set `enable_debug = true` in both the `[hypervisor.qemu]` and `[proxy.kata]` sections
+- For `use_vsock = true`:
+    - Set `enable_debug = true` in both the `[hypervisor.qemu]` and `[shim.kata]` sections
+
+For generic information on enabling debug in the confiuration file, see the
```
gnawux: spell check is not satisfied with the "confiuration" :joy:

grahamwhaley:
fixed a typo, spin again.

Issue #596, opened 2020-02-18
add notes about how to enable debug for shimv2


lifupan:
# Description of problem

For shimv2 + containerd combination, not only the kata's debug will impact how the kata runtime/agent log, but the containerd's debug option will also do impact it. we should notify 
this in doc.

# Expected result

(replace this text with an explanation of what you thought would happen)

# Actual result

(replace this text with details of what actually happened)

PR #597, opened 2020-02-18
debug: add notes about how to enable debug for shimv2

lifupan:
The containerd's debug option will determine whether
the kata's log forared to containerd's log pipe or
not.

Fixes:#596

Signed-off-by: fupan.lfp <fupan.lfp@antfin.com>

### DIFF
--- a/Developer-Guide.md
+++ b/Developer-Guide.md
@@ -6,6 +6,7 @@
     * [Check hardware requirements](#check-hardware-requirements)
     * [Configure to use initrd or rootfs image](#configure-to-use-initrd-or-rootfs-image)
     * [Enable full debug](#enable-full-debug)
+        * [debug logs and shimv2](#debug-logs-and-shimv2)
         * [journald rate limiting](#journald-rate-limiting)
             * [systemd-journald suppressing messages](#systemd-journald-suppressing-messages)
             * [Disabling systemd-journald rate limiting](#disabling-systemd-journald-rate-limiting)
@@ -140,6 +141,21 @@ $ sudo install -o root -g root -m 0640 /usr/share/defaults/kata-containers/confi
 $ sudo sed -i -e 's/^# *\(enable_debug\).*=.*$/\1 = true/g' /etc/kata-containers/configuration.toml
 $ sudo sed -i -e 's/^kernel_params = "\(.*\)"/kernel_params = "\1 agent.log=debug initcall_debug"/g' /etc/kata-containers/configuration.toml
 ```
+
+### debug logs and shimv2
+
+If you are using `containerd` and Kata shimv2 to launch `kata-runtime`, `containerd`'s debug option 
+will determine if the `kata-runtime`'s log is forwarded to `containerd`'s log pipe. If you want to see 
+the Kata's log (including the `runtime` and `agent` logs) in the `containerd` log, then you need to 
+enable `containerd`'s debug by adding the following to your `containerd`'s config file:
+
+```toml
+[debug]
+        level = "debug"
+```
+
+For much info about the containerd's debug, please see the
+[`containerd` documentation](https://github.com/containerd/containerd/blob/master/docs/getting-started.md).
 
 ### journald rate limiting
 


grahamwhaley:
Thanks @lifupan - that's looking good, useful!. I've edited it a bit (see below) to try and make it flow and pass the CI. I suspect the CI 'magic spellchecker' is still going to complain, but we can work on that. I put it into a new subsection as it looked strange as one big note section. wdyt?

----
### Debug logs and `shimv2`

If you are using `containerd` and Kata shimv2 to launch `kata-runtime`, `containerd`'s debug option will determine if the `kata-runtime`'s log is forwarded to `containerd`'s log pipe. If you want to see the Kata's log (including the `runtime` and `agent` logs) in the `containerd` log, then you need to enable `containerd`'s debug by adding the following to your `containerd`'s config file:

```toml
[debug]
        level = "debug"
```

For much info about the containerd's debug, please see the [`containerd` documentation](https://github.com/containerd/containerd/blob/master/docs/getting-started.md).

----

Raw markdown looks like:

````markdown
### Debug logs and `shimv2`

If you are using `containerd` and Kata shimv2 to launch `kata-runtime`, `containerd`'s debug option 
will determine if the `kata-runtime`'s log is forwarded to `containerd`'s log pipe. If you want to see 
the Kata's log (including the `runtime` and `agent` logs) in the `containerd` log, then you need to 
enable `containerd`'s debug by adding the following to your `containerd`'s config file:

```toml
[debug]
        level = "debug"
```

For much info about the containerd's debug, please see the
[`containerd` documentation](https://github.com/containerd/containerd/blob/master/docs/getting-started.md).
````

**Comment by lifupan** [comment]
@grahamwhaley Thanks, i had updated it. Please take a look.

**Comment by grahamwhaley** [comment]
(/me hijacks closed PR...)
Hi @lifupan - I was looking at how containerd does logs, and how kata fits in with that. I found that [containerd provides a log pipe for shims to write to](https://github.com/containerd/containerd/blob/master/runtime/v2/README.md#logging) that is created in the namespace of the shim, and if I go look on the node where a kata container is running, I see a pipe called `log`.... 

```
$ kubectl describe pod nginx-deployment-qemu-7cd654c9fc-l2vbx
...
Controlled By:  ReplicaSet/nginx-deployment-qemu-7cd654c9fc
Containers:
  nginx:
    Container ID:   containerd://9fe3b90eedd7ff16edff0cbb5e4b2b297876faab671eaf745a5e40ea907a2b07

#[now on the node]
$ pwd
/run/containerd/io.containerd.runtime.v2.task/k8s.io/9fe3b90eedd7ff16edff0cbb5e4b2b297876faab671eaf745a5e40ea907a2b07
$ ls -la
total 20
drwx--x--x 3 root root  140 Feb 25 11:16 .
drwx--x--x 6 root root  120 Feb 25 11:16 ..
-rw-r--r-- 1 root root   98 Feb 25 11:16 address
-rw-r--r-- 1 root root 4365 Feb 25 11:16 config.json
prwx------ 1 root root    0 Feb 25 11:16 log
```

I had a look at the kata shimv2 code, and I don't think it makes use or has the option/facility to make use of that `log` pipe - do we think maybe it should?

**Comment by grahamwhaley** [comment]
@lifupan - I tried to dig into this some more - I wanted to try and work out how to tie the shimv2 logs into fluentd, to go with my pending https://github.com/kata-containers/documentation/pull/602, because I think it will be quite different when using shimv2....

I found that containerd looks to have a separate shim_debug setting, so if we do:

```toml
[plugins.linux]
  shim_debug=true
```

in the containerd config file, that also gets us the shimv2 (kata) debug logs - but without afaict turning on the rest of the containerd debug. That feels like we should add that to the above document as well. Code is at https://github.com/containerd/containerd/blob/master/runtime/v1/linux/runtime.go#L408-L412 afaict:

```go
		if r.config.ShimDebug {
			go copyAndClose(os.Stderr, shimStderrLog)
		} else {
			go copyAndClose(ioutil.Discard, shimStderrLog)
		}
```

and the same for stdout.

afaict, there is no way to enable just getting kata non-debug logs with shimv2 (like, I presume we'd like warnins, errors and probably info to be coming out in a log somewhere...), which is a little sad :-(. It feels maybe a little wrong to have to enable something called 'debug' in order to get kata logs.
/cc @gnawux @amshinde for any thoughts. We (I) should probably take this to another thread if we need to discuss more...

boy, trying to work out how the logs and debugging works though the twisty maze of containerd, shims, configs, logrus and stuff is ... tedious....

amshinde:
@grahamwhaley If the warning and error logs are not getting propogated without setting the debug flag, I think it is worth raising an issue for this in the containerd repo and try to get their input on this.

grahamwhaley:
@amshinde - I've dug in and found more info now - opened a local Issue at https://github.com/kata-containers/runtime/issues/2513 to discuss.
I'll separately update our docs to clarify that there are two runtime debug settings in the containerd config file that affect us (I just need to double check that...)

grahamwhaley:
> @grahamwhaley If the warning and error logs are not getting propogated without setting the debug flag, I think it is worth raising an issue for this in the containerd repo and try to get their input on this.

For completeness, I should clarify this. I was wrong (the code is a little twisty and some things in shimv2 default to different levels than classic kata). I believe we do get shimv2 logs in containerd logs, but we default to 'Warn' level rather than 'Info' level. See https://github.com/kata-containers/runtime/issues/2513 for more details.

PR #607, opened 2020-03-05
Dev: shimv2 logs: Elaborate on shimv2 log enabling

grahamwhaley:
There are two 'debug' settings in the containerd config file that
affect the shimv2 runtime log output. Add the other method to the
existing documentation, and also note that enabling full containerd
debug also affects all of containerd.

The commit also re-generates the TOC, which seems to correct a
few anomolies there.

Fixes: #596


grahamwhaley:
green green green...

PR #643, opened 2020-04-14
Changed one letter

Bryan1269:
docker run --runtime=kata-nemu -it busybox -> docker run --runtime=kata-qemu -it busybox
I tried to run Kata with VirtioFS but found that this may be a possible typo.

I read the daemon.json generated while update Kata and none printed "nemu". But correct me if I'm wrong.

Regards,
Bryan

P.S. : This is my first time contributing to an open source project

### DIFF
--- a/how-to/how-to-use-virtio-fs-with-kata.md
+++ b/how-to/how-to-use-virtio-fs-with-kata.md
@@ -39,7 +39,7 @@ This will place the Kata release artifacts in `/opt/kata`, and update Docker's c
 
 Once installed, start a new container, utilizing NEMU + `virtiofs`:
 ```bash
-$ docker run --runtime=kata-nemu -it busybox
+$ docker run --runtime=kata-qemu -it busybox
 ```
 
 Verify the new container is running with the NEMU hypervisor as well as using `virtiofsd`. To do this look for the hypervisor path and the `virtiofs` daemon process on the host:


grahamwhaley:
Thanks @Bryan1269 - well, there is/was a `nemu` as well as a `qemu` - so, at the time I suspect it was not a typo... but, I also think nemu is somewhat deprecated, so maybe it would make sense to update that.... wdyt @amshinde @gnawux ?

grahamwhaley:
@Bryan1269 - I'll also point you at our [documentation for pull request formatting and requirements](https://github.com/kata-containers/community/blob/master/CONTRIBUTING.md#patch-format), as the CI's are noting you have some bits missing, like the sign off and reference Issue number etc.

how-to/how-to-use-virtio-fs-with-kata.md
```diff
@@ -39,7 +39,7 @@ This will place the Kata release artifacts in `/opt/kata`, and update Docker's c
 
 Once installed, start a new container, utilizing NEMU + `virtiofs`:
 ```bash
-$ docker run --runtime=kata-nemu -it busybox
+$ docker run --runtime=kata-qemu -it busybox
```
amshinde, line 42: Thanks @Bryan1269. `kata-nemu` is deprecated and this needs to be updated. But the runtime to be used here is `kata-qemu-virtiofs` instead of `kata-qemu`, as this runtime makes use of qemu with virtio-fs support.

egernst:
Looks like this was fixed upstream.  Thanks for the contribution @bryan1269

PR #711, opened 2020-08-26
Update installation information

fidencio:
This PR aims to update the installation information for the distros claimed to be supported.

The main changes here are:
- CentOS:  use packages from the virtualization special interest group;
- Fedora: use the packages provided by the distro;
- openSUSE: use the packages provided by the distro;
- RHEL: drop the information;
- SLES: use the packages provided by the distro;

I still have to check what's the state of the packages on Ubuntu and Debian, once it's done I can remove the "WIP:" tag.

PR #737, opened 2020-09-24
Update Supported Distributions table and Fedora installation.

willianrampazzo:
Update the Supported Distributions table with what is currently built at http://download.opensuse.org/repositories/home:/katacontainers:/

Include the command to install the Fedora distro packages.

Signed-off-by: Willian Rampazzo <willianr@redhat.com>


willianrampazzo:
I did not see #711 before I send this. That PR is much more complete. In this case, closing this.

PR #783, opened 2021-01-22
Glossary of terms for Kata Containers.

hspease99:
I added a glossary to support SEO tactics. The terms added are: Kata
Conatiners, container software, container runtime interface, virtual
machine software, container virtualization, container security solutions,
serverless containers, pod containers, virtual machine monitor, private cloud,
infrastructure architecture, public cloud, and auto scaling.

Signed-Off-By: Helena Spease helena@openstack.org


jodh-intel (commented):
Hi @hspease99 - thanks for raising this!

A couple of thoughts:

- This repo is for legacy (Kata 1.x) documentation, so a PR should probably be raised in the 2.x repo instead (since the 2.x docs live at https://github.com/kata-containers/kata-containers/tree/main/docs).
- That said, I'm actually wondering if it would make more sense to turn this into a wiki page at this stage since there are lot of other terms that could be added I think. After we've got lots of input from folks on the wiki page, we could then switch it to an "in-tree" doc on a PR like this.
- There is a complication though: 1.x has a number of components that are no longer used in 2.x (proxy and shim processes for example) so we may need to ponder how best to deal with this.

In terms of other entries, areas to look at are:

- http://www.katacontainers.io/
- https://github.com/kata-containers/documentation/wiki/UserGuide
- https://github.com/kata-containers/kata-containers/blob/main/docs (in particular https://github.com/kata-containers/kata-containers/blob/main/docs/design/architecture.md)
- The data files for our custom spell-checker which we use to spell check all documentation PRs (https://github.com/kata-containers/tests/tree/main/cmd/check-spelling, specifically https://github.com/kata-containers/tests/tree/main/cmd/check-spelling/data)
---

Related to this: we're thinking of having a monthly "docs call" to discuss all things documentation and forge a plan for how to improve on what we already provide. More details will appear on the public ML soon I hope... ;)

cmaf (commented):
@hspease99 I think a glossary would be extremely helpful and really hope this moves forward! 

I agree with @jodh-intel, this should be raised in https://github.com/kata-containers/kata-containers/.

